PR #1, opened 2016-09-09
Fix bugs in lb_vpx_vip and lb_vpx_service

minsikl:
SoftLayer_Network_Application_Delivery_Controller service provides Netscaper VPX management services. createLiveLoadBalancer, deleteLiveLoadBalancer, deleteLiveLoadBalancerService, and updateLiveLoadBalancer functions only accept one request in one Netscaler VPX at the same time. It is necessary to handle exceptions to manage concurrent requests. 

Loop and exception handling logics are added to create, update, delete functions in lb_vpx_vip and lb_vpx_service. Managed errors are as follows:
- "Operation already in progress" : Retry in 10 secs.
- "Unable to find object with unknown..." : Complete the delete operation.
- "Internal Error" : Retry in 10 secs.
- "No Service" : Retry in 10 secs. 

Test files are also updated to fix wrong test scenarios.

### DIFF
--- a/softlayer/resource_softlayer_lb_vpx_service.go
+++ b/softlayer/resource_softlayer_lb_vpx_service.go
@@ -14,6 +14,7 @@ import (
 	"github.com/softlayer/softlayer-go/services"
 	"github.com/softlayer/softlayer-go/session"
 	"github.com/softlayer/softlayer-go/sl"
+	"time"
 )
 
 func resourceSoftLayerLbVpxService() *schema.Resource {
@@ -119,7 +120,7 @@ func resourceSoftLayerLbVpxServiceCreate(d *schema.ResourceData, meta interface{
 	}
 
 	// Check if there is an existed loadbalancer service which has same name.
-	log.Printf("[INFO] Validating LoadBalancer Service Name %s", *lb_services[0].Name)
+	log.Printf("[INFO] Creating LoadBalancer Service Name %s validation", *lb_services[0].Name)
 
 	_, err = network.GetNadcLbVipServiceByName(sess, nadcId, vipName, serviceName)
 	if err == nil {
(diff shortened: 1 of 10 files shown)


softlayer/resource_softlayer_lb_vpx_service.go
```diff
@@ -129,7 +130,20 @@ func resourceSoftLayerLbVpxServiceCreate(d *schema.ResourceData, meta interface{
 
 	log.Printf("[INFO] Creating LoadBalancer Service %s", *lb_services[0].Name)
 
-	successFlag, err := service.Id(nadcId).UpdateLiveLoadBalancer(lbVip)
+	successFlag := true
+	for count := 0; count < 10; count++ {
+		successFlag, err = service.Id(nadcId).UpdateLiveLoadBalancer(lbVip)
+		log.Printf("[INFO] Creating LoadBalancer Service %s successFlag : %t", *lb_services[0].Name, successFlag)
```
renier: small suggestion to store the service name early above and reuse throughout for readability

``` go
serviceName := *lb_services[0].Name
```

softlayer/resource_softlayer_lb_vpx_service.go
```diff
@@ -202,14 +217,34 @@ func resourceSoftLayerLbVpxServiceUpdate(d *schema.ResourceData, meta interface{
...
+		log.Printf("[INFO] Updating Loadbalancer service %s successFlag : %t", serviceName, successFlag)
+
+		if err != nil && strings.Contains(err.Error(), "Operation already in progress") {
+			log.Printf("[INFO] Updating Loadbalancer service %s Error : %s. Retry in 10 secs", serviceName, err.Error())
+			time.Sleep(time.Second * 10)
+			continue
+		}
+
+		break
+	}
```
renier: this block of code seems to be the same as in the Create function. If so, it would be good to refactor into its own function and reuse from both places.

PR #2, opened 2016-09-11
Fix dns resources

renier:
All dns tests pass now.

Decided to remove record management from the dns domain resource. Was getting too unmanageable with SoftLayer automatically creating the NS records for you under the covers. However, added a _faked_ attribute called "target" in which you specify the primary ip address that the domain will resolve to. This is similar to how it works in the portal. This enables the domain create to succeed, since it will include an A record with the target ip address. Updates to this target record are supported through the domain resource.

All other records can be created using the domain record resource. SRV record creation and editing was tested.

Other things were improved:
- Preventing users from managing NS records. These are automatically created by SoftLayer with correct values. Users should never have to manage these themselves. Trying to manage these result in conflicts with changes SoftLayer can do to the record properties
- Preventing users from using uppercase letters in IPv6 addresses. SoftLayer lowercases IPv6 addresses in the record's data without warning. This becomes unmanageable if the address in .TF uses uppercase letters (which is valid in IPv6) as Terraform will always see a change necessary after an apply.

### DIFF
--- a/softlayer/resource_softlayer_dns_domain.go
+++ b/softlayer/resource_softlayer_dns_domain.go
@@ -1,6 +1,7 @@
 package softlayer
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"strconv"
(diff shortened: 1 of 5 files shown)


softlayer/resource_softlayer_dns_domain.go
```diff
@@ -197,45 +93,77 @@ func resourceSoftLayerDnsDomainRead(d *schema.ResourceData, meta interface{}) er
...
 func resourceSoftLayerDnsDomainUpdate(d *schema.ResourceData, meta interface{}) error {
-	// TODO - update is not supported - implement delete-create?
-	return fmt.Errorf("Not implemented. Update Dns Domain is currently unsupported")
+	// If the target has been updated, find the corresponding dns record and update its data.
+	sess := meta.(*session.Session)
+	domainId, _ := strconv.Atoi(d.Id())
+
+	if !d.HasChange("target") {
+		return nil
+	}
```
riethm, line 125: This check probably isn't necessary in the first place, since `target` is the only field for which the schema allows an update (all other fields are either computed or `ForceNew: true`).  

But if this is to guard against some unforeseen logic, or simply to highlight the fact that only `target` can be updated, then it may be appropriate to at least return a non-nil `error`.  Otherwise, this is essentially a silent fail.

softlayer/resource_softlayer_dns_domain_record.go
```diff
@@ -178,30 +274,55 @@ func resourceSoftLayerDnsDomainRecordRead(d *schema.ResourceData, meta interface
...
+		d.Set("refresh", *result.Refresh)
+	}
+
+	if result.ResponsiblePerson != nil {
+		d.Set("responsible_person", *result.ResponsiblePerson)
+	}
+
+	if result.Retry != nil {
+		d.Set("retry", *result.Retry)
+	}
```
riethm: Do we need to consider that any of the above fields could have changed from non-nil to nil, in which case we should update the resource to reflect the new `nil` value?

softlayer/resource_softlayer_dns_domain.go
renier: I wasn't 100% sure about when update would be called. You are saying that with how the properties are configured now, Update will never be called unless there is already a change in target?

softlayer/resource_softlayer_dns_domain_record.go
renier: you suggesting this? `d.Set("refresh", sl.Get(result.Refresh, nil))`

PR #3, opened 2016-09-15
Feature/vlan

minsikl:
Provides softlayer_network_vlan resource. 3 files are added as follows:
1. docs/softlayer_network_vlan.md
2. softlayer/resource_softlayer_network_vlan.go
3. softlayer/resource_softlayer_network_vlan_test.go

vendor directory has been updated to provide a new helper function hardware.GetRouterByName()


softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+
+func resourceSoftLayerNetworkVlanCreate(d *schema.ResourceData, meta interface{}) error {
+	sess := meta.(*session.Session)
+	router := d.Get("primary_router_hostname").(string)
+	name := d.Get("name").(string)
+
+	vlanType := d.Get("type").(string)
+	if vlanType != "PRIVATE" && vlanType != "PUBLIC" {
+		return fmt.Errorf("Error creating network vlan: vlanType should be either 'PRIVATE' or 'PUBLIC'")
+	}
```
renier: I think this is better done as a validate function on the schema.

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+
+	vlan, err := service.Id(vlanId).Mask(
+		"id," +
+			"name," +
+			"primaryRouter[datacenter[name]]," +
+			"primaryRouter[hostname]," +
+			"vlanNumber," +
+			"billingItem[recurringFee]," +
+			"guestNetworkComponentCount," +
+			"subnets[networkIdentifier,cidr,subnetType]",
```
renier: Performance nit. This is better to have as some constant globally in the package.

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+
+	d.Set("id", *vlan.Id)
+	d.Set("vlan_number", *vlan.VlanNumber)
+	d.Set("child_resource_count", *vlan.GuestNetworkComponentCount)
+	if vlan.Name != nil {
+		d.Set("name", *vlan.Name)
+	} else {
+		d.Set("name", "")
+	}
```
renier: `d.set("name", sl.Get(vlan.Name, ""))`

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+	} else {
+		d.Set("softlayer_managed", false)
+	}
+
+	// Subnets
+	subnets := make([]map[string]interface{}, 0)
+
+	for _, elem := range vlan.Subnets {
+		subnet := make(map[string]interface{})
+		subnet["subnet"] = *elem.NetworkIdentifier + "/" + strconv.Itoa(*elem.Cidr)
```
renier: it seems to me that using fmt.Sprintf would perform better.

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+	vlanId, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return fmt.Errorf("Not a valid vlan ID, must be an integer: %s", err)
+	}
+
+	// Check if the VLAN is existed.
+	_, err = service.Id(vlanId).Mask("id").GetObject()
+	if err != nil {
+		return fmt.Errorf("Error deleting Network Vlan: %s", err)
+	}
```
renier: Don't think we need to check if the vlan existed separately. I think the next call to the api below will tell you that already. If it existed, this results in two api calls, rather than just one.

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+
+	billingItem, err := service.Id(vlanId).GetBillingItem()
+	if err != nil {
+		return fmt.Errorf("Error deleting Network Vlan: %s", err)
+	}
+	if billingItem.Id == nil {
+		return nil
+	}
```
renier: should add a comment here based on our discussion

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+
+	success, err := services.GetBillingItemService(sess).Id(*billingItem.Id).CancelService()
+	if err != nil {
+		return err
+	}
+
+	if !success {
+		return fmt.Errorf("SoftLayer reported an unsuccessful cancellation")
+	}
```
renier: Not sure. I think we should ignore success? and just return `err` back.

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+		return false, fmt.Errorf("Not a valid vlan ID, must be an integer: %s", err)
+	}
+
+	_, err = service.Id(vlanId).Mask("id").GetObject()
+
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
```
renier: here, you can do this at line 283: `return err == nil, err`

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+
+func findNetworkVlanByOrderId(sess *session.Session, orderId int) (datatypes.Network_Vlan, error) {
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"pending"},
+		Target:  []string{"complete"},
+		Refresh: func() (interface{}, string, error) {
+			vlans, err := services.GetAccountService(sess).
+				Filter(filter.Build(
+					filter.Path("networkVlans.billingItem.orderItem.order.id").
+						Eq(strconv.Itoa(orderId)))).
```
renier:
``` go
Filter(
        filter.Path("networkVlans.billingItem.orderItem.order.id").
        Eq(strconv.Itoa(orderId)).
        Build()
)
```

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+
+	if len(datacenter) > 0 {
+		dc, err = location.GetDatacenterByName(sess, datacenter, "id")
+		if err != nil {
+			return &datatypes.Container_Product_Order_Network_Vlan{}, err
+		}
+	} else {
+		return &datatypes.Container_Product_Order_Network_Vlan{},
+			fmt.Errorf("datacenter name is empty.")
+	}
```
renier: Would rewrite it as:

``` go
if datacenter == "" {
        return &datatypes.Container_Product_Order_Network_Vlan{},
                     fmt.Errorf("datacenter name is empty.")
}

dc, err := location.GetDatacenterByName(sess, datacenter, "id")
if err != nil {
    return &datatypes.Container_Product_Order_Network_Vlan{}, err
}
```

Then would not have to pre-declare dc and err above.

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+		rt, err = hardware.GetRouterByName(sess, router, "id")
+		productOrderContainer.RouterId = rt.Id
+		if err != nil {
+			return &datatypes.Container_Product_Order_Network_Vlan{},
+				fmt.Errorf("Error creating network vlan: %s", err)
+		}
+	}
+
+	return &productOrderContainer, nil
+}
```
renier: @riethm Would this function make a good candidate for adding as a helper to softlayer-go?

renier (changes requested):

docs/softlayer_network_vlan.md
```diff
@@ -0,0 +1,55 @@
+#### `softlayer_network_vlan`
```
riethm: Just want to put it out there - Are we sure we want to name this `softlayer_network_vlan` instead of perhaps shortening to just `softlayer_vlan` ?

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+				ForceNew: true,
+			},
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"primary_router_hostname": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
riethm: Not sure I understand the implication of a property being declared both `Computed` and `Optional`.  You can always change a property's value programmatically, whether it is declared as `Computed` or not.  So this seems to be effectively the same as just making the property `Optional` only (`Computed` has no additional effect)?

softlayer/resource_softlayer_network_vlan.go
riethm: @renier Seems like there is potential here for a helper.  For consideration: would the proposed helper just return the order container, or place the order (`OrderNetworkVlan()`) and return a receipt?

softlayer/resource_softlayer_network_vlan.go
```diff
@@ -0,0 +1,419 @@
...
+				{
+					Id: subnetItems[0].Prices[0].Id,
+				},
+			},
+			Quantity: sl.Int(1),
+		},
+	}
+
+	if len(router) > 0 {
+		rt, err = hardware.GetRouterByName(sess, router, "id")
```
riethm: It took me a few minutes to understand why the `GetRouterByName` helper was placed into a `hardware` package.  I do see why now, but still wonder if anyone would think to look there.  Maybe we should consider an alias function in another package (network?)  Just a thought - nothing to address for this PR.

softlayer/resource_softlayer_network_vlan.go
riethm: 👍

softlayer/resource_softlayer_network_vlan.go
riethm: Does this logic work?  i.e., If Terraform receives a non-nil error from `...Exists`, does it even consider the first (`bool`) return value, or would the error disqualify the `exists` test?

Thinking it also doesn't handle non-404 errors (500, network, etc.)

docs/softlayer_network_vlan.md
renier: 👍

softlayer/resource_softlayer_network_vlan.go
renier: _helper/schema/resource.go_ says it checks the error first:

``` go
                exists, err := r.Exists(data, meta)
                if err != nil {
                        return s, err
                }
                if !exists {
                        return nil, nil
                }
```

you are saying we should check for 404 if we have an error?

``` go
if err != nil && err.(sl.Error).StatusCode == 404 {
        return false, nil
}

return err == nil, err
```

?

softlayer/resource_softlayer_network_vlan.go
minsikl: Actually, I couldn't find other examples which use both Computed and Optional, but here is a test result of `Optional` property only.

_vlan.tf_

```
resource "softlayer_network_vlan" "test_vlan" {
   name = "test_vlan2"
   datacenter = "lon02"
   type = "PUBLIC"
   primary_subnet_size = 8
}
```

_terraform_ _show_ (after _terraform_ _apply_)

```
softlayer_network_vlan.test_vlan:
  id = 1369721
  child_resource_count = 0
  datacenter = lon02
  name = test_vlan2
  primary_router_hostname = fcr01a.lon02
  primary_subnet_size = 8
  softlayer_managed = false
  subnets.# = 1
  subnets.1384167489.subnet = 159.8.147.208/29
  subnets.1384167489.subnet_type = ADDITIONAL_PRIMARY
  type = PUBLIC
  vlan_number = 1255
```

_terraform_ _plan_

```
softlayer_network_vlan.test_vlan
primary_router_hostname: "fcr01a.lon02" => ""
```

With current codes, terraform tries to delete primary_router_hostname whenever I execute `terraform apply`

softlayer/resource_softlayer_network_vlan.go
riethm: Right.  I think that would work.

softlayer/resource_softlayer_network_vlan.go
riethm: I see.  That makes sense.  And because it's `ForceNew: true`, we never have to deal with the case where the config is updated.  Nice.

softlayer/resource_softlayer_network_vlan.go
minsikl: Right. At first, I thought that Hardware or Hardware_Router services would provide those functions. However, GetRouterByName does not use any Hareware_\* services. Only a return type of GetRouterByName is `Hardware`. Except Router, other resources such as SoftLayer_Network_Pod and SoftLayer_Network_Backbone are using Network_\* datatypes. So their helper functions will be located in network package in the future. So network package can be good place for GetRouterByName.

docs/softlayer_network_vlan.md
minsikl: `softlayer_vlan` is better I think.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Agree. Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Not sure. But I couldn't find any cases in my test. I deleted that logic.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

softlayer/resource_softlayer_network_vlan.go
minsikl: Applied.

docs/softlayer_network_vlan.md
minsikl: So change the resource name to `softlayer_vlan`?

minsikl: Updated.

renier (approved):

PR #4, opened 2016-09-20
Make virtual guest resource property names clearer

renier:
- Change image to os_reference_code
- Change block_device_template_group_gid to image_id
- Change public_network_speed to network_speed
- Tell terraform that image_id and os_reference_code conflict with each
  other.
- Fix other intellij/go warnings.
- Update tests (they pass)
- Update documentation.


riethm:
👍

PR #6, opened 2016-09-22
Feature/new shared local lb

riethm:
Enable support for shared local load balancers.

Added `dedicated` field to `softlayer_lb_local` schema.
- true: dedicated
- false: shared (default)

For shared local LBs, SSL enablement will be determined by the presence or absence of a `security_certificate_id` in the user's configuration.

Updated docs, acceptance tests, and vendored SDK


renier:
Looks good to me, but would like Minsik to take a look.

softlayer/resource_softlayer_lb_local.go
```diff
@@ -91,14 +103,28 @@ func resourceSoftLayerLbLocalCreate(d *schema.ResourceData, meta interface{}) er
...
 	} else {
-		keyname = "LOAD_BALANCER_DEDICATED_WITH_SSL_OFFLOAD"
+		categoryCode = product.ProxyLoadBalancerCategoryCode
+		if _, ok := d.GetOk("security_certificate_id"); ok {
+			d.Set("ssl_enabled", true)
+			keyFormatter = "LOAD_BALANCER_%d_VIP_CONNECTIONS_WITH_SSL_OFFLOAD"
+		} else {
+			d.Set("ssl_enabled", false)
+			keyFormatter = "LOAD_BALANCER_%d_VIP_CONNECTIONS"
+		}
```
minsikl, line 128: `ha_enabled` is a required field and shared load balancers don't support HA. It is necessary to return an error if `dedicated = false` and `ha_enabled = true`  in a create function.

minsikl:
Question: Is there any way to send `{"parameters":[{"securityCertificateId":null}]}` in update function? When I create a shared loadbalancer with `security_certificate_id` and remove it from the .tf configuration and execute `terraform apply` again, the terraform tries to update `security_certificate_id=0` and returns an error.

minsikl:
It looks like that some functions have to be implemented for load balancer as follows:
1) SSL offload management - function to provide enable or disable SSL certificate for local load balancer. 
2) # of connections upgrade function for shared load balancer.

They don't have to be implemented in this PR.

riethm:
(number of) `connections` is `ForceNew: true` so I don't think we have to address the upgrade scenario.  Other points seem valid.  Will look into them and respond to each if they can be addressed.

softlayer/resource_softlayer_lb_local.go
riethm: Done.  See 402377e

I added validation around the `ha_enabled` flag, to ensure an error is reported if it is set when creating a shared load balancer.  I also made the schema field optional, with default value of `false`, since:
1. it seemed a sensible default
2. it spares the user creating a shared load balancer from having to explicitly specify a property that is not applicable.  

The field is still `ForceNew: true`

Let me know if you have any concerns with this change, and we can just revert it to as it was before, with the new validation being the only change.

riethm:
@minsikl 8842f22 enables removal of an SSL cert by passing `null` instead of 0 in the EditObject call.

riethm:
@minsikl All above issues relevant to enabling shared local LBs have now been addressed.  

As we have been discussing, SSL still needs some work (a lot actually) - it was never properly done.  Issue is that just adding an SSL cert to a VIP does not automatically activate SSL acceleration on a load balancer.  Additional methods have to be called (stop/startSsl()), but a number of factors need to be taken into account:
1. has a cert been loaded to the VIP?
2. Is SSL already active/inactive (API method hangs if you try to activate SSL where it is already active)
3. When considering activation, are there any service groups which use HTTPS as their routing type (startSsl() call fails if there are no HTTPS service groups)
4. Similarly, when deleting a service group, need to check that it is the last SG using SSL - otherwise this could leave the LB in an inconsistent state that will require manual recovery.

We've also discovered that the way our local LB and related schemas are defined, there is the distinct possibility for race conditions when multiple service groups are being created at once, all accessing  a common state object (ssl state).

Given the amount of rework that will be required to fully support automatic SSL activation/deactivation, I think we can raise these as a separate issue to be worked on in an upcoming release.

minsikl:
The code looks good.

PR #8, opened 2016-09-23
Feature/datasource ssh keys

riethm:
- Added a datasource `softlayer_ssh_key`, which can be used to import ssh keys into the local state, using `label` as the import key
- Removed the `ssh_key_labels` property from the softlayer_virtual_guest resource schema
- Renamed `name` to `label` in the softlayer_ssh_key resource schema, to reflect the name of the field as is commonly visible to users in the portal or CLI (`name` is an API field)


docs/datasources/softlayer_ssh_key.md
```diff
@@ -0,0 +1,41 @@
+# `softlayer_ssh_key`
+
+Use this data source to import the details of an *existing* SSH key as a read-only data source.
```
renier: I would add that you can use this to get the SSH key id that you could use in other resources like virtual_guest.

docs/datasources/softlayer_ssh_key.md
```diff
@@ -0,0 +1,41 @@
...
+
+The fields of the data source can then be referenced by other resources within the same configuration using
+interpolation syntax. For example, when specifying SSH keys in a softlayer_virtual_guest resource configuration,
+the numeric "IDs" are often unknown. Using the above data source as an example, it would be possible to
+reference the `id` property in a softlayer_virtual_guest resource:
+
+```hcl
+resource "softlayer_virtual_guest" "vm1" {
+    ...
+    ssh_keys = ["${data.softlayer_ssh_key.public_key.id}"]
```
renier, line 21: will this result in a string or an int?

docs/datasources/softlayer_ssh_key.md
```diff
@@ -0,0 +1,41 @@
...
+}
+```
+
+## Argument Reference
+
+* `label` - (Required) The label of the SSH key, as it was defined in SoftLayer
+* `most_recent` - (Optional) If more than SSH key matches the label, use the most recent key
+
+NOTE: If more or less than a single match is returned by the search, Terraform will fail.
+Ensure that your search is specific enough to return a single SSH key only,
```
renier: Ensure that your label...

docs/datasources/softlayer_ssh_key.md
```diff
@@ -0,0 +1,41 @@
...
+or use most_recent to choose the most recent one.
```
renier: or use `*most_resent*`...

docs/softlayer_virtual_guest.md
```diff
@@ -118,10 +118,9 @@ The following arguments are supported:
     * *Default*: nil
     * *Optional*
     * *Conflicts with* `ssh_key_labels`
```
renier: Need to remove this line, too.

docs/datasources/softlayer_ssh_key.md
riethm: Good question - I have never been able to get it to work with quotes omitted.  Here's what I get when I use this (quoted) syntax:

```
ssh_keys.#:               "1"
ssh_keys.0:               "383111"
```

So clearly, string.  But `ssh_keys = [383111]` yields the exact same result.  Maybe Terraform stores everything as `interface{}`, displays as strings, and type asserts as needed.

softlayer/resource_softlayer_ssh_key.go
```diff
@@ -28,11 +28,17 @@ func resourceSoftLayerSSHKey() *schema.Resource {
...
 				Type:     schema.TypeString,
 				Required: true,
 			},
 
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				Removed:  "Use 'label' instead",
+			},
+
```
renier: I would not bother with deprecations... Up to you. and it's nice that they have _Removed_ for these situations, but I don't think we are there yet to need it.

softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -187,17 +187,16 @@ func resourceSoftLayerVirtualGuest() *schema.Resource {
...
 			"ssh_key_labels": {
-				Type:          schema.TypeList,
-				Optional:      true,
-				Elem:          &schema.Schema{Type: schema.TypeString},
-				ConflictsWith: []string{"ssh_keys"},
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Removed:  fmt.Sprintf("Use the interpolation method described at %s", datasourceSshKeyDoc),
 			},
```
renier: This has been there for such a short time, I'd just remove.

renier (commented):
I think we need to evolve our docs directory. Looking at the terraform project, they have the provider documentation, then two subdirectories. One for resources, another for data sources. I would like to see us emulate that.

docs/datasources/softlayer_ssh_key.md
renier: ugh, and `ssh_keys = [${data.softlayer_ssh_key.public_key.id}]` would not work then? That means I made a mistake in some of the doc updates.

docs/datasources/softlayer_ssh_key.md
riethm: Fixed

docs/datasources/softlayer_ssh_key.md
riethm: Fixed

docs/softlayer_virtual_guest.md
riethm: Fixed

softlayer/resource_softlayer_ssh_key.go
riethm: Removed `Removed`

softlayer/resource_softlayer_virtual_guest.go
riethm: Removed `Removed`

PR #10, opened 2016-09-24
New image template data source

renier:


softlayer/data_source_softlayer_image_template_test.go
```diff
@@ -0,0 +1,39 @@
...
+					resource.TestCheckResourceAttr(
+						"data.softlayer_image_template.tfacc_img_tmpl",
+						"name",
+						"jumpbox",
+					),
+					resource.TestCheckResourceAttr(
+						"data.softlayer_image_template.tfacc_img_tmpl",
+						"id",
+						"988251",
+					),
```
riethm, line 27: Suggestion: just check that some ID was stored, so that the test case is more portable.  e.g.:

`resource.TestMatchResourceAttr("data.softlayer_image_template.tfacc_img_tmpl", "id", regexp.MustCompile("^[0-9]+")),`

PR #11, opened 2016-09-30
User Resource Enhancements

riethm:
- Schema fields "timezone" and "user_status" now accept string values instead of integer-based IDs.
- Logic to handle permission changes uses set math to determine exactly which permissions to add/remove, versus just wiping all permissions and bulk loading the new set.
- Removed a TODO related to password update.  There is no current plan to support updating a user's password via Terraform, as the process cannot be fully automated.

### DIFF
--- a/docs/resources/softlayer_user.md
+++ b/docs/resources/softlayer_user.md
@@ -26,8 +26,7 @@ resource "softlayer_user" "joe" {
         "RESET_PORTAL_PASSWORD"
     ]
     state        = "GA"
-    timezone     = 114
-    user_status  = 1001
+    timezone     = "EST"
     username     = "joedoe"
 }
 ```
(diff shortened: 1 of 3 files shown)


docs/resources/softlayer_user.md
```diff
@@ -26,8 +26,7 @@ resource "softlayer_user" "joe" {
         "RESET_PORTAL_PASSWORD"
     ]
     state        = "GA"
-    timezone     = 114
-    user_status  = 1001
+    timezone     = "EST"
```
renier, line 29: Cool

softlayer/resource_softlayer_user.go
```diff
@@ -404,16 +429,22 @@ func resourceSoftLayerUserUpdate(d *schema.ResourceData, meta interface{}) error
 }
 
 func resourceSoftLayerUserDelete(d *schema.ResourceData, meta interface{}) error {
-	service := services.GetUserCustomerService(meta.(*session.Session))
+	sess := meta.(*session.Session)
+	service := services.GetUserCustomerService(sess)
 
 	id, _ := strconv.Atoi(d.Id())
 
+	userCancelStatusID, err := getUserStatusIDByName(sess, userCustomerCancelStatus)
```
renier: Is it likely for this pending cancel status ID to change? It could save roundtrips if we just have a constant for the ID.

softlayer/resource_softlayer_user.go
```diff
@@ -431,3 +462,41 @@ func resourceSoftLayerUserExists(d *schema.ResourceData, meta interface{}) (bool
...
+	}
+
+	return -1, fmt.Errorf("Timezone %s could not be found", shortName)
+
+}
+
+func getUserStatusIDByName(sess *session.Session, name string) (int, error) {
+	statuses, err := services.GetUserCustomerStatusService(sess).
+		Mask("id,keyName").
+		GetAllObjects()
```
renier, line 483: Let me guess. Filtering doesn't work here either...

riethm: hah.  Nope.

softlayer/resource_softlayer_user.go
riethm: Good point.  I have not heard of it changing.  Will use a constant

PR #12, opened 2016-10-03
[WIP - do not merge] Feature/globalip

riethm:


riethm:
@srinivasdandibhotla Opening this PR so that we can comment on the changes as development progresses

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+        if err != nil {
+                return fmt.Errorf("Not a valid global ip ID, must be an integer: %s", err)
+        }
+
+        globalIp, err := service.Id(globalIpId).Mask(GlobalIpMask).GetObject()
+        if err != nil {
+                if strings.Contains(err.Error(), "404 Not Found") {
+                        d.SetId("")
+                        return nil
+                }
```
riethm: @srinivasdandibhotla String comparison is not a reliable way to check for a 404.  Instead, type-assert the error to an `sl.Error` and check the status code directly, for example:

``` go
if apiErr, ok := err.(sl.Error); ok && apiErr.StatusCode == 404 {
        d.SetId("")
        return nil
    }
```

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+                return fmt.Errorf("Not a valid global ip ID, must be an integer: %s", err)
+        }
+        
+        if d.HasChange("routes_to") {
+                _, err = service.Id(globalIpId).Route(sl.String(d.Get("routes_to").(string)))
+        }        
+
+        if err != nil {
+                return fmt.Errorf("Error editing Global Ip: %s", err)
+        }
```
riethm: This works, but would it be more clear if we check for the error in the `if d.HasChange() {}` block?

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+                return fmt.Errorf("Not a valid global ip ID, must be an integer: %s", err)
+        }
+
+        billingItem, err := service.Id(globalIpId).GetBillingItem()
+        if err != nil {
+                return fmt.Errorf("Error deleting global ip: %s", err)
+        }
+
+        if billingItem.Id == nil {
+                return nil
```
riethm: silent fail?

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+        sess := meta.(*session.Session)
+        service := services.GetNetworkSubnetIpAddressGlobalService(sess)
+
+        globalIpId, err := strconv.Atoi(d.Id())
+        if err != nil {
+                return false, fmt.Errorf("Not a valid ID, must be an integer: %s", err)
+        }
+
+        result, err := service.Id(globalIpId).GetObject()
+        return result.Id != nil && err == nil && *result.Id == globalIpId, nil
```
riethm: @srinivasdandibhotla The problem with the above is that, _if_ there is an error, we are suppressing it by returning `nil` for the second return value.

Instead, when there is an error preventing a conclusive existence check, we need to report it to Terraform by returning that error (optionally wrapped), so that Terraform will not incorrectly mark an active resource as destroyed, or conversely, report a destroyed resource as active.  So we need to separate the error check and existence check, and ensure any error preventing the check is reported.

We have an issue in our internal tracker to address this for other resources.  Let's use the correct way for this new resource.

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+
+        if err != nil {
+                return datatypes.Network_Subnet_IpAddress_Global{}, err
+        }
+
+        var result, ok = pendingResult.(datatypes.Network_Subnet_IpAddress_Global)
+
+        if ok {
+                return result, nil
+        }
```
riethm: You could combine the above lines, e.g.:

``` go
if result, ok := pendingResult.(datatypes.Network_Subnet_IpAddress_Global); ok {
    return result, nil
}
```

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+
+        return datatypes.Network_Subnet_IpAddress_Global{},
+                fmt.Errorf("Cannot find global ip with order id '%d'", orderId)
+}
+
+func buildGlobalIpProductOrderContainer(d *schema.ResourceData, sess *session.Session, packageType string) (
+        *datatypes.Container_Product_Order_Network_Subnet, error) {
+        // hard coding the data center
+        datacenter := "sjc03"
```
riethm: Assuming this is temporary

riethm:
@srinivasdandibhotla I don't see the changes to provider.go here
Also, I assume you are still working on acceptance tests

softlayer/resource_softlayer_global_ip.go
```diff
@@ -158,6 +156,9 @@ func resourceSoftLayerGlobalIpExists(d *schema.ResourceData, meta interface{}) (
         }
 
         result, err := service.Id(globalIpId).GetObject()
+        if err != nil {
+                return false, fmt.Errorf("Error retrieving global ip: %s", err)
+        }
         return result.Id != nil && err == nil && *result.Id == globalIpId, nil
```
riethm: You can remove `err == nil` from this check now that it is checked above

renier:
@srinivasdandibhotla You'll need to do a rebase:

```
git fetch upstream # assuming that is what you named the remote
git rebase -i upstream/master (from your feature/globalip branch)
```

riethm:
@srinivasdandibhotla Please run `make fmt` and commit the formatted code.

riethm:
When I run the acceptance test, I get: 

`SoftLayer_Exception_Public: You are not allowed to order additional Global IPv4 addresses. Please contact Sales. (HTTP 500)`

I have opened a ticket to find out why Global IPs are not being purged once they are cancelled.

srinivasdandibhotla:
Did git fetch and rebase. Then committed the formatted code though provider.go is still a conflicting file.  This file has only a single line change "softlayer_global_ip":              resourceSoftLayerGlobalIp(),

riethm:
Srinivas: You need to rebase against the 'upstream' repository (softlayer/terraform-provider-softlayer).  Check your remotes, and add the upstream repository as a remote:

```
git remote -v show
git remote add upstream git@github.com:softlayer/terraform-provider-softlayer.git
```

Then follow Renier's instructions to fetch the upstream and rebase

riethm:
@srinivasdandibhotla SoftLayer cleared up some of the old Global IPs so I was able to run the test.  It failed with a nil pointer dereference:

```
2016/10/06 09:03:31 [DEBUG] Path:  https://api.softlayer.com/rest/v3/SoftLayer_Network_Subnet_IpAddress_Global/50303.json?objectMask=mask%5Bid%2CipAddress%5BipAddress%5D%2CdestinationIpAddress%5BipAddress%5D%5D
2016/10/06 09:03:31 [DEBUG] Parameters:
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0xb0 pc=0x9ab22]

goroutine 53 [running]:
panic(0x8f11a0, 0xc82000e110)
    /usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/softlayer/terraform-provider-softlayer/softlayer.resourceSoftLayerGlobalIpRead(0xc820286600, 0x8b9260, 0xc82026e600, 0x0, 0x0)
    /Users/riethm/Projects/go/terraform-softlayer/src/github.com/softlayer/terraform-provider-softlayer/softlayer/resource_softlayer_global_ip.go:101 +0x7f2
github.com/softlayer/terraform-provider-softlayer/softlayer.resourceSoftLayerGlobalIpUpdate(0xc820286600, 0x8b9260, 0xc82026e600, 0x0, 0x0)
    /Users/riethm/Projects/go/terraform-softlayer/src/github.com/softlayer/terraform-provider-softlayer/softlayer/resource_softlayer_global_ip.go:122 +0x670
```

Pointing to this line of code in `resourceSoftLayerGlobalIpRead()` (L101):

```
    d.Set("routes_to", *globalIp.DestinationIpAddress.IpAddress)
```

When I ran the API call mentioned in the error, `DestinationIpAddress` was populated:

``` json
{
  "id": 50303,
  "destinationIpAddress": {
    "ipAddress": "119.81.82.163"
  },
  "ipAddress": {
    "ipAddress": "169.55.61.84"
  }
}
```

So it seems that the destinationIpAddress might not be set immediately after `Route()` returns.  Possibly we need to put a wait on it?

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,265 @@
...
+        d.Set("id", *globalIp.Id)
+        d.Set("ip_address", *globalIp.IpAddress.IpAddress)
+        d.Set("routes_to", *globalIp.DestinationIpAddress.IpAddress)        
+
+        return nil
+}
+
+func resourceSoftLayerGlobalIpUpdate(d *schema.ResourceData, meta interface{}) error {
+        sess := meta.(*session.Session)
+        service := services.GetNetworkSubnetIpAddressGlobalService(sess)
```
Freyert: Are we assuming the underlying API layer will handle errors appropriately?

Would it be better if we explicitly handled any errors coming from this call here? Rather than throwing an error when unable to convert a string to an integer.

renier: @Freyert Line 111 does not involve a call to the API. Is that what you are referring to? Or what do you mean?

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,250 @@
...
+	globalIpId, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return fmt.Errorf("Not a valid global ip ID, must be an integer: %s", err)
+	}
+
+	if d.HasChange("routes_to") {
+		_, err = service.Id(globalIpId).Route(sl.String(d.Get("routes_to").(string)))
+		if err != nil {
+			return fmt.Errorf("Error editing Global Ip: %s", err)
+		}
```
minsikl: It is necessary to check an active transaction. SoftLayer_Network_Subnet_IpAddress_Global::getActiveTransaction can be used to check the active transaction.

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,279 @@
...
+			d.SetId("")
+			return nil
+		}
+
+		return fmt.Errorf("Error retrieving Global Ip: %s", err)
+	}
+
+	d.Set("id", *globalIp.Id)
+	d.Set("ip_address", *globalIp.IpAddress.IpAddress)
+	d.Set("routes_to", *globalIp.DestinationIpAddress.IpAddress)
```
minsikl: Some cases, `globalIp.DestinationIpAddress` can be `null`. It is necessary to check if `globalIp.DestinationIpAddress` is null or not.

softlayer/resource_softlayer_global_ip_test.go
```diff
@@ -0,0 +1,116 @@
...
+		*globalIp = foundGlobalIp
+
+		return nil
+	}
+}
+
+const testAccCheckSoftLayerGlobalIpConfig_basic = `
+resource "softlayer_global_ip" "test-global-ip" {
+    routes_to = "119.81.82.163"
+}`
```
minsikl, line 111: Please remove hardcoding information "119.81.82.163". You can create 2 VMs and use their public IP addresses as `routes_to`.

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,279 @@
...
+func resourceSoftLayerGlobalIpUpdate(d *schema.ResourceData, meta interface{}) error {
+	sess := meta.(*session.Session)
+	service := services.GetNetworkSubnetIpAddressGlobalService(sess)
+
+	globalIpId, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return fmt.Errorf("Not a valid global ip ID, must be an integer: %s", err)
+	}
+
+	if d.HasChange("routes_to") {
```
renier: Since _Update_ is called from _Create_, I think this condition will be false that first call (`d.HasHcange("routes_to")` will be false). So either pull out the code within the condition as a helper function that can be called from _Create_ without any interruptions and can allso be called from here (_Update_). Or, remove the `if d.HasChange("routes_to")` conditional, since _routes_to_ is the only updatable field, so _Update_ won't be called unless it has changed.

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,279 @@
...
+		if err != nil {
+			return fmt.Errorf("Error waiting for global ip destination ip address to become active: %s", err)
+		}
+
+		if _, ok := pendingResult.(datatypes.Network_Subnet_IpAddress_Global); ok {
+			return nil
+		}
+	}
+
+	return resourceSoftLayerGlobalIpRead(d, meta)
```
renier, line 150: Don't think we should call _Read_ from _Update_. I know that it is to compute the ip address field upon Creation, since _Update_ is called from _Create_ due to how _routes_to_ gets set after the fact. Instead, I think that  `findGlobalIpByOrderId` should return a GlobalIP that already has the _ip address_ set. That way we can compute (`d.Set()`) it from _Create_, and thus _Update_ doesn't need to call _Read_ anymore.

Otherwise _Read_ would always be called on subsequent _Update_s unnecessarily.

renier:
This is being continued in #36

PR #13, opened 2016-10-03
Vet all

renier:
Make `vet` a required part of building the binary. This also fixes intellij warnings on some related files.

### DIFF
--- a/softlayer/resource_softlayer_scale_group.go
+++ b/softlayer/resource_softlayer_scale_group.go
@@ -245,7 +245,7 @@ func buildScaleVlansFromResourceData(d *schema.Set, meta interface{}) ([]datatyp
 
 		if len(networkVlans) < 1 {
 			return nil, fmt.Errorf(
-				"Unable to locate the specified vlan: %d.%s",
+				"Invalid vlan: could not find specified vlan with number %d and router %s.",
 				vlanNumber, primaryRouterHostname)
 		}
 
--- a/softlayer/resource_softlayer_scale_group.go
+++ b/softlayer/resource_softlayer_scale_group.go
@@ -245,9 +245,8 @@ func buildScaleVlansFromResourceData(d *schema.Set, meta interface{}) ([]datatyp
 
 		if len(networkVlans) < 1 {
 			return nil, fmt.Errorf(
-				"Unable to locate a vlan matching the provided router hostname and vlan number: %s/%s",
-				primaryRouterHostname,
-				vlanNumber)
+				"Unable to locate the specified vlan: %d.%s",
+				vlanNumber, primaryRouterHostname)
 		}
 
 		scaleNetworkVlans = append(
@@ -683,5 +682,5 @@ func getLocationGroupRegionalId(sess *session.Session, locationGroupRegionalName
 		}
 	}
 
-	return -1, fmt.Errorf("Invalid regional_group_id:", locationGroupRegionalName)
+	return -1, fmt.Errorf("Invalid regional_group_id: %s", locationGroupRegionalName)
 }
--- a/softlayer/resource_softlayer_vlan.go
+++ b/softlayer/resource_softlayer_vlan.go
@@ -1,6 +1,7 @@
 package softlayer
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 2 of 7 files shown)


softlayer/resource_softlayer_scale_group.go
```diff
@@ -245,9 +245,8 @@ func buildScaleVlansFromResourceData(d *schema.Set, meta interface{}) ([]datatyp
 
 		if len(networkVlans) < 1 {
 			return nil, fmt.Errorf(
-				"Unable to locate a vlan matching the provided router hostname and vlan number: %s/%s",
-				primaryRouterHostname,
-				vlanNumber)
+				"Unable to locate the specified vlan: %d.%s",
+				vlanNumber, primaryRouterHostname)
```
riethm, line 249: Accidentally part of this commit?

riethm: Edit: I missed the relevant change (%s -> %d). I'm still not sure about squashing vlan number and hostname together. Thought that was for a future PR.

softlayer/resource_softlayer_vlan.go
```diff
@@ -37,65 +38,65 @@ func resourceSoftLayerVlan() *schema.Resource {
...
 				Required: true,
 				ForceNew: true,
 			},
-			"type": &schema.Schema{
+			"type": {
 				Type:     schema.TypeString,
 				Required: true,
 				ForceNew: true,
-				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+				ValidateFunc: func(v interface{}, k string) (ws []string, errs []error) {
```
riethm, line 54: Wow

softlayer/resource_softlayer_scale_group.go
renier: This is just squashing it for the purposes of the error message. It was squashing it before in the message as `%s/%s` (router hostname first) anyway. I think the dot should be a more recognizable delimiter for the squashed format.

softlayer/resource_softlayer_vlan.go
renier: Otherwise, it conflicts with the package name `errors`, and then it gets weird when using `errors.new()` in the same function space.

riethm: right.  I saw that.  Wow = good catch

softlayer/resource_softlayer_scale_group.go
riethm: Yes, but in 1) removing the field names and 2) choosing a delimiter that appears in your data 100% of the time, it's now ambiguous.  It may be more recognizable to us, but there is no such thing as vlan '1234.bcr01a.sjc03'.  It isn't a thing, and a user new to SoftLayer will find this confusing.

In the past, this combined form was actually displayed in the UI.  SL split them out, probably for good reason (portal vs API representation).

renier: @riethm so what would be an improvement over `bcr01a.sjc03/1234` as it was before? Since that can't be any less ambiguous or confusing to the type of users you pose.

riethm: Ok.  I see your point now.  It's not very clear as-is either.

I believe this is somewhat moot right now anyway, as we were planning to re-work this schema to accept a VLAN ID (once the new datasource is in-place).  So this method will probably be moved to the new vlan datasource anyway?  That said, if we really want to eliminate all ambiguity, then I guess the message would need to look something like:
"Invalid vlan: the combination of vlan number %d and hostname %s do not identify an existing vlan"

renier: 06afd1e

renier: > It may be more recognizable to us, but there is no such thing as vlan '1234.bcr01a.sjc03'. It isn't a thing, and a user new to SoftLayer will find this confusing.

You are right. I just got the order wrong :-)
screen_shot_2016-10-03_at_10_23_21_pm

riethm: Interesting.  I probably shouldn't point out that the hostname is written before the number here.

renier: That's the order I got wrong. I used to see it with the number first.

PR #14, opened 2016-10-04
New Bare Metal resource

renier:
I have named some properties different than what you'll find in the virtual guest resource. This is on purpose. I feel these are better names and intend to update the virtual guest in the same way.

### DIFF
--- a/softlayer/resource_softlayer_bare_metal.go
+++ b/softlayer/resource_softlayer_bare_metal.go
@@ -336,19 +336,35 @@ func resourceSoftLayerBareMetalDelete(d *schema.ResourceData, meta interface{}) 
 		return fmt.Errorf("Not a valid ID, must be an integer: %s", err)
 	}
 
-	_, err = WaitForNoActiveTransactions(d, meta)
-
+	_, err = WaitForNoBareMetalActiveTransactions(id, meta)
 	if err != nil {
-		return fmt.Errorf("Error deleting bare metal server, couldn't wait for zero active transactions: %s", err)
+		return fmt.Errorf("Error deleting bare metal server while waiting for zero active transactions: %s", err)
 	}
 
 	_, err = service.Id(id).DeleteObject()
-
 	if err != nil {
 		return fmt.Errorf("Error deleting bare metal server: %s", err)
 	}
 
 	return nil
+}
+
+func resourceSoftLayerBareMetalExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	service := services.GetHardwareService(meta.(*session.Session))
+
+	id, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return false, fmt.Errorf("Not a valid ID, must be an integer: %s", err)
+	}
+
+	result, err := service.Id(id).GetObject()
+	if err != nil {
+		if apiErr, ok := err.(sl.Error); !ok || apiErr.StatusCode != 404 {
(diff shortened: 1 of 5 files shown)


renier:
@riethm This was actually not ready yet this morning, but it is now.

softlayer/resource_softlayer_bare_metal.go
```diff
@@ -0,0 +1,442 @@
...
+	d.Set("network_speed", *result.PrimaryNetworkComponent.MaxSpeed)
+	if result.PrimaryIpAddress != nil {
+		d.Set("public_ipv4_address", *result.PrimaryIpAddress)
+	}
+	d.Set("private_ipv4_address", *result.PrimaryBackendIpAddress)
+
+	d.Set("private_network_only", *result.PrivateNetworkOnlyFlag)
+	d.Set("hourly_billing", *result.HourlyBillingFlag)
+
+	if result.PrimaryNetworkComponent.NetworkVlan != nil {
```
riethm, line 315: Have you confirmed that `PrimaryNetworkComponent` is non-nil for a private-only BM, or should we guard against nil for this field as well?

softlayer/resource_softlayer_bare_metal.go
```diff
@@ -0,0 +1,442 @@
...
+		d.Set("public_vlan_id", *result.PrimaryNetworkComponent.NetworkVlan.Id)
+	}
+
+	if result.PrimaryBackendNetworkComponent.NetworkVlan != nil {
+		d.Set("private_vlan_id", *result.PrimaryBackendNetworkComponent.NetworkVlan.Id)
+	}
+
+	userData := result.UserData
+	if len(userData) > 0 && userData[0].Value != nil {
+		d.Set("user_data", *userData[0].Value)
```
riethm: `d.Set("user_metadata"...` ?

softlayer/resource_softlayer_bare_metal.go
```diff
@@ -0,0 +1,442 @@
...
+
+	return stateConf.WaitForState()
+}
+
+func WaitForNoBareMetalActiveTransactions(id int, meta interface{}) (interface{}, error) {
+	log.Printf("Waiting for server (%d) to have zero active transactions", id)
+	service := services.GetHardwareServerService(meta.(*session.Session))
+
+	stateConf := &resource.StateChangeConf{
+		Pending: []string{"", "active"},
```
riethm: I only see `""` returned when there is an error retrieving the object.  But since the refresh func returns the error, I don't think this state is ever checked:

``` go
// ... in WaitForState()
// If the Refresh function returns a error, exit immediately with that error.
```

net: I don't think you need `""` here.  Or, if you intend to retry in the event of an error, then don't return the error, and return a state listed here (`""` or `"pending"`)

softlayer/resource_softlayer_bare_metal.go
renier: No. For VMs, PrimaryNetworkComponent is never nil for private-only VMs (the fields of the components may be nil), but I have not tried private BMs.

renier:
FYI, One of the travis-ci checks failed because this branch doesn't have a .travis.yml file yet.

softlayer/resource_softlayer_bare_metal.go
renier: Tested private network only. The read function worked fine.

Issue #19, opened 2016-10-07
Creating multiple (>2) lb_local_service results in timeouts, putting terraform in inconsistent state


iankendrick:
When terraform several multiple lb_local_service objects simultaneously, some of the creations will time-out. Following apply, Softlayer seems to have correct state, but terraform has incorrect IP for the timed-out lb_local_service objects. Refreshing and re-running plan does not correct the issue, and tf state file must be manually edited.
### Steps to reproduce:

Run terraform file to create 5 servers, load balancer, service group and related services [1]
### Expected result:

Virtual guests are created, alongside a Load Balancer and Service Group. Service Group contains 5 Local Services, correspond to the 5 Virtual guests.
### Actual Result

Time-out creating anywhere from 1 to 3 of 5 Local Services [2]. Neither `terraform refresh` or `terraform plan` reflects correct state. `terraform apply` will always fail [3].
1. terraform file:
   
   ```
   resource "softlayer_virtual_guest" "test" {
       name = "test-0${count.index}"
       count = 5
       domain = "test.com"
       ssh_keys = [######]
       os_reference_code = "UBUNTU_14_64"
       datacenter = "tor01"
       network_speed = 100
       cpu = 4
       ram = 1024
   }
   
   # Create a new local load balancer
   resource "softlayer_lb_local" "test_lb_local" {
       connections = 250
       datacenter = "tor01"
       ha_enabled = false
       dedicated = false
   }
   
   # Create a new local load balancer service
   resource "softlayer_lb_local_service" "test_lb_local_service" {
       count = 5
       ip_address_id = "${element(softlayer_virtual_guest.test.*.ip_address_id_private, count.index)}"
       port = 31443
       enabled = true
       service_group_id = "${softlayer_lb_local_service_group.test_service_group.service_group_id}"
       weight = 1
       health_check_type = "TCP"
   }
   
   
   # Create a new local load balancer service group
   resource "softlayer_lb_local_service_group" "test_service_group" {
       port = 443
       routing_method = "CONSISTENT_HASH_IP"
       routing_type = "TCP"
       load_balancer_id = "${softlayer_lb_local.test_lb_local.id}"
       allocation = 100
   }
   ```
2. error message:
   
   ```
    softlayer_lb_local_service.test_lb_local_service.2: Creation complete
    softlayer_lb_local_service.test_lb_local_service.1: Creation complete
    softlayer_lb_local_service.test_lb_local_service.3: Creation complete
[...]

PR #20, opened 2016-10-07
Local Load Balancer Service fixes

riethm:
- Fixed a bad filter which was being ignored, causing all service objects to be returned instead of just the matching one, in turn causing multiple services to be assigned the same ID
- Added WaitForState() logic around create/update/delete calls, to prevent timeouts leading to unrecoverable state inconsistencies
- Sync "enabled" flag during refresh

Addresses #19

### DIFF
--- a/softlayer/resource_softlayer_lb_local_service.go
+++ b/softlayer/resource_softlayer_lb_local_service.go
@@ -132,8 +132,8 @@ func resourceSoftLayerLbLocalServiceCreate(d *schema.ResourceData, meta interfac
 		Id(sgID).
 		Mask("mask[id,port,ipAddressId]").
 		Filter(filter.New(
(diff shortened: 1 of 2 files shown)


softlayer/resource_softlayer_lb_local_service.go
```diff
@@ -208,18 +228,37 @@ func resourceSoftLayerLbLocalServiceUpdate(d *schema.ResourceData, meta interfac
...
+			}
+
+			return true, "complete", nil
+		},
+		Timeout:    10 * time.Minute,
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	_, err = stateConf.WaitForState()
```
renier: seems like you could put this in a function so that it can be reused from create and update.

PR #26, opened 2016-10-09
Provisioning VMs over XMLRPC doesn't work

renier:
This brings in fixes I made from the xmlrpc library (and subsequently to softlayer-go with regards to xmlrpc _omitempty_). It also fixes problems with some api calls timing out within the _Wait_ functions found during testing of VM provisioning over XMLRPC.

There was also an odd issue with the private router hostname not being returned upon _CreatObject_. Happened when testing a private-only VM over XMLRPC, but cannot be sure this does not happen with public VMs as well. Fix is intermixed into commit 46f6acf. This issue will be avoided altogether once we move to specifying VLAN ids in VM resources only.


riethm:
@renier Under vendor/github.com/softlayer/softlayer-go there are two new files I'm pretty sure you didn't intend to include:

main.go
test.go

Otherwise, this looks okay to me.

renier:
@riethm Thanks for catching that. I've removed those files from the history.

PR #27, opened 2016-10-09
Add data source for VLAN

renier:


softlayer/data_source_softlayer_vlan.go
```diff
@@ -0,0 +1,74 @@
...
+	if name != "" {
+		networkVlans, err := service.
+			Mask("id").
+			Filter(filter.Path("networkVlans.name").Eq(name).Build()).
+			GetNetworkVlans()
+		if err != nil || len(networkVlans) == 0 {
+			return fmt.Errorf("Error obtaining VLAN id: %s", err)
+		}
+
+		d.SetId(fmt.Sprintf("%d", *networkVlans[0].Id))
```
minsikl: `number` and `router_hostname` have to be updated.

softlayer/data_source_softlayer_vlan.go
```diff
@@ -0,0 +1,74 @@
...
+	} else if number != 0 && routerHostname != "" {
+		id, err := getVlanId(number, routerHostname, meta)
+		if err != nil {
+			return fmt.Errorf("Error obtaining VLAN id: %s", err)
+		}
+
+		d.SetId(fmt.Sprintf("%d", id))
```
minsikl: `name` has to be updated.

softlayer/data_source_softlayer_vlan.go
```diff
@@ -0,0 +1,74 @@
...
+				Type:          schema.TypeInt,
+				Optional:      true,
+				ConflictsWith: []string{"name"},
+			},
+
+			"router_hostname": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"name"},
+			},
```
minsikl, line 46: I think that consistent schema names should be used for data and resources. Field names `number` and `router_houstname` are different from the field names in a vlan resource.

minsikl: How about other computed fields such as `datacenter`, `type`, `primary_subnet_size`, `softlayer_managed`, `child_resource_count`, and `subnets`? Especially, `subnets` can be used for provisioning automation.

softlayer/data_source_softlayer_vlan.go
```diff
@@ -0,0 +1,74 @@
...
+	number := d.Get("number").(int)
+	routerHostname := d.Get("router_hostname").(string)
+
+	if name != "" {
+		networkVlans, err := service.
+			Mask("id").
+			Filter(filter.Path("networkVlans.name").Eq(name).Build()).
+			GetNetworkVlans()
+		if err != nil || len(networkVlans) == 0 {
+			return fmt.Errorf("Error obtaining VLAN id: %s", err)
```
minsikl: service.Mask().Filter().GetNetworkVlans() returns `[]` without errors if it does not find any vlan. `err` in fmt.Errorf can be a `nil`. Error messages have to be separated as follows:

```
        if err != nil {
            return fmt.Errorf("Error obtaining VLAN id: %s", err)
        } else if len(networkVlans) == 0 {
            return fmt.Errorf("Error obtaining VLAN id: No VLAN with the name '%s'", name)
        }
```

softlayer/data_source_softlayer_vlan.go
renier: Not sure @minsikl, will secondary_router_hostname ever be something we need to add to the vlan resource or its corresponding data source? Hoping we can simplify down to router_hostname. For now, I can specify in the docs that we mean primary.

About the other fields. I can work on adding subnets to the data source, if that will be useful for informing other resources. The others, I feel we should wait until there is a need to add it.

softlayer/data_source_softlayer_vlan.go
renier: Ok, it would print as `%!s(<nil>)` on the output.

softlayer/data_source_softlayer_vlan.go
renier: Why? these are not computed fields. The idea is to get the VLAN id which will be used in the other resoruces (lb_vpx, virtual_guest, bare_metal, scale_group). But what resource will be in need of number/router_hostname from providing the name in the data source?

softlayer/data_source_softlayer_vlan.go
renier: Same question as above for the name from providing number/router_hostname.

softlayer/data_source_softlayer_vlan.go
renier: Adding a note, to update to `router_hostname` in softlayer_vlan resource.

renier: Adding a note, to add subnets.

PR #29, opened 2016-10-14
Supress diff on vlan ids list (scale group)

renier:
Appreciate help testing this @minsikl when you can.

### DIFF
--- a/softlayer/resource_softlayer_scale_group.go
+++ b/softlayer/resource_softlayer_scale_group.go
@@ -128,33 +128,11 @@ func resourceSoftLayerScaleGroup() *schema.Resource {
 			},
 
 			"network_vlan_ids": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeInt},
-				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					o := strings.TrimRight(strings.TrimLeft(old, " ["), "] ")
-					n := strings.TrimRight(strings.TrimLeft(new, " ["), "] ")
-					oldIds := strings.Split(o, ",")
-					newIds := strings.Split(n, ",")
-					if len(oldIds) != len(newIds) {
-						return false
-					}
-
-					for _, newId := range newIds {
-						found := false
-						for _, oldId := range oldIds {
-							if strings.TrimSpace(newId) == strings.TrimSpace(oldId) {
-								found = true
-								break
-							}
-						}
-
-						if !found {
-							return false
-						}
-					}
-
-					return true
+				Set: func(v interface{}) int {
+					return v.(int)
 				},
 			},
 		},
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_scale_group.go
```diff
@@ -359,19 +384,11 @@ func resourceSoftLayerScaleGroupRead(d *schema.ResourceData, meta interface{}) e
...
-	// is created. Else, this code needs to be refactored to use a TypeSet.
-	if vlanTotal == 0 {
-		vlanIds := make([]int, vlanTotal)
-		for _, vlan := range slGroupObj.NetworkVlans {
-			vlanIds = append(vlanIds, *vlan.Id)
-		}
-		d.Set("network_vlan_ids", vlanIds)
+	vlanIds := make([]int, len(slGroupObj.NetworkVlans))
+	for i, vlan := range slGroupObj.NetworkVlans {
+		vlanIds[i] = *vlan.Id
```
minsikl: It is necessary to change `*vlan.Id` to `*vlan.networkVlanId`. `SoftLayerScaleGroupObjectMask` also should contains `"networkVlans[networkVlanId,..."`.

softlayer/resource_softlayer_scale_group.go
```diff
@@ -131,6 +131,31 @@ func resourceSoftLayerScaleGroup() *schema.Resource {
 				Type:     schema.TypeList,
 				Optional: true,
 				Elem:     &schema.Schema{Type: schema.TypeInt},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					o := strings.TrimRight(strings.TrimLeft(old, " ["), "] ")
```
minsikl: `o` only contains one element of an array. For example, if `network_vlans_ids = [1001,1002,1003,1004]` in tf file and terraform's read function returns `[1004,1003,1002,1001]`, DiffSuppressfunc will be invoked 4 times as follows:
1st call : o = 1001, n = 1004
2nd call : o = 1002, n = 1003
3rd call : o = 1003, n = 1002
4th call : o = 1004, n = 1001

renier: As discussed in the scrum, we can try to address this by simply checking whether `n` is already a member of `d.Get('network_vlan_ids')`. If true, we suppress, otherwise we don't suppress. This assumes that `d` has the old snapshot of the resource data, and not the new.

renier:
@minsikl This is ready to be merged.

softlayer/resource_softlayer_scale_group.go
```diff
@@ -490,17 +486,19 @@ func resourceSoftLayerScaleGroupUpdate(d *schema.ResourceData, meta interface{})
...
-				if n == o {
+		for _, o := range oldIds {
+			for _, n := range newIds {
+				if n.(int) == o.(int) {
 					goto nextOld
 				}
 			}
 
-			_, err = scaleNetworkVlanService.Id(o).DeleteObject()
+			_, err = scaleNetworkVlanService.Id(o.(int)).DeleteObject()
```
minsikl: Need to use scale vlan ID as follows:

```
targetVlans, err = scaleGroupService.Id(groupId).Filter(filter.Build(
                filter.Path("networkVlans.networkVlanId").
                    Eq(strconv.Itoa(o.(int))))).GetNetworkVlans()
...
_, err = scaleNetworkVlanService.Id(*targetVlans[0].Id).DeleteObject()
```

PR #32, opened 2016-10-19
Fix scalegroup

minsikl:
- Fixed virtual guest template fields
- Changed virtual_guest_id, port, and health_check options to optional fields.
- Updated documentation of scale_gruop.md.

### DIFF
--- a/softlayer/resource_softlayer_scale_group.go
+++ b/softlayer/resource_softlayer_scale_group.go
@@ -411,9 +411,9 @@ func populateMemberTemplateResourceData(template datatypes.Virtual_Guest) map[st
 	}
 
 	if template.BlockDeviceTemplateGroup != nil {
-		d["block_device_template_group_gid"] = sl.Get(template.BlockDeviceTemplateGroup.GlobalIdentifier)
+		d["image_id"] = sl.Get(template.BlockDeviceTemplateGroup.GlobalIdentifier)
 	} else {
-		d["block_device_template_group_gid"] = nil
+		d["image_id"] = nil
 	}
 
 	if len(template.UserData) > 0 {
--- a/softlayer/resource_softlayer_scale_group.go
+++ b/softlayer/resource_softlayer_scale_group.go
@@ -386,7 +386,7 @@ func populateMemberTemplateResourceData(template datatypes.Virtual_Guest) map[st
 	d["name"] = *template.Hostname
 	d["domain"] = *template.Domain
 	d["datacenter"] = *template.Datacenter.Name
-	d["public_network_speed"] = *template.NetworkComponents[0].MaxSpeed
+	d["network_speed"] = *template.NetworkComponents[0].MaxSpeed
 	d["cpu"] = *template.StartCpus
 	d["ram"] = *template.MaxMemory
 	d["private_network_only"] = *template.PrivateNetworkOnlyFlag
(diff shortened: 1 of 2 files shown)


softlayer/resource_softlayer_scale_group.go
```diff
@@ -309,6 +301,36 @@ func resourceSoftLayerScaleGroupCreate(d *schema.ResourceData, meta interface{})
...
+		isLoadBalancerEmpty = false
+		loadBalancers[0].Port = sl.Int(port.(int))
+	}
+
+	if isLoadBalancerEmpty {
+		return []datatypes.Scale_LoadBalancer{}, nil
+	} else {
+		return loadBalancers, nil
+	}
+}
```
renier, line 332: Would `len(loadBalancers) == 0` serve the same purpose as `isLoadBalancerEmpty`?

softlayer/resource_softlayer_scale_group.go
```diff
@@ -334,29 +356,31 @@ func resourceSoftLayerScaleGroupRead(d *schema.ResourceData, meta interface{}) e
...
-	if *healthCheckObj.Type.Keyname == HEALTH_CHECK_TYPE_HTTP_CUSTOM {
-		for _, elem := range healthCheckObj.Attributes {
-			switch *elem.Type.Keyname {
-			case "HTTP_CUSTOM_TYPE":
-				currentHealthCheck["custom_method"] = *elem.Value
-			case "LOCATION":
-				currentHealthCheck["custom_request"] = *elem.Value
-			case "EXPECTED_RESPONSE":
-				currentHealthCheck["custom_response"] = *elem.Value
+	if slGroupObj.LoadBalancers != nil && len(slGroupObj.LoadBalancers) > 0 {
```
renier: `.LoadBalancers` is a slice, not a pointer nor interface{}. So I don't think it would ever be nil.

softlayer/resource_softlayer_scale_group.go
```diff
@@ -411,9 +435,9 @@ func populateMemberTemplateResourceData(template datatypes.Virtual_Guest) map[st
 	}
 
 	if template.BlockDeviceTemplateGroup != nil {
-		d["block_device_template_group_gid"] = sl.Get(template.BlockDeviceTemplateGroup.GlobalIdentifier)
+		d["image_id"] = sl.Get(template.BlockDeviceTemplateGroup.GlobalIdentifier)
 	} else {
-		d["block_device_template_group_gid"] = nil
+		d["image_id"] = nil
```
renier, line 440: I think you can also pass `nil` as the second argument to sl.Get() to get it as the default value

softlayer/resource_softlayer_scale_group.go
minsikl: `len(loadBalancers)` is `1` because `loadBalancers := []datatypes.Scale_LoadBalancer{{}}`

softlayer/resource_softlayer_scale_group.go
minsikl: Applied.

softlayer/resource_softlayer_scale_group.go
minsikl: Yeah, however I think that I still need to use `if` clause to check if `template.BlockDeviceTemplateGroup` is nil or not.

renier: Ok. I was wondering if that would not be the default already, if there was no key in the map for image_id.

softlayer/resource_softlayer_scale_group.go
renier: I see.

PR #34, opened 2016-10-21
Fix the read function in scale_group

minsikl:
The following bugs are fixed.
1) virtual_guest_member_template has TypeList type. populateMemberTemplateResourceData
should return []map[string]interface{} instead of map[string]interface{}
2) frontend_vlan_id and backend_vlan_id are changed to public_vlan_id and private_vlan_id
in populateMemberTemplateResourceData function.
3) Else clauses are removed in populateMemberTemplateResourceData function.
4) Fixed network_vlan part of read function in scale_group.

### DIFF
--- a/softlayer/resource_softlayer_scale_group.go
+++ b/softlayer/resource_softlayer_scale_group.go
@@ -396,14 +396,13 @@ func resourceSoftLayerScaleGroupRead(d *schema.ResourceData, meta interface{}) e
 		}
 		d.Set("network_vlan_ids", vlanIds)
 	}
-
 	virtualGuestTemplate := populateMemberTemplateResourceData(*slGroupObj.VirtualGuestMemberTemplate)
 	d.Set("virtual_guest_member_template", virtualGuestTemplate)
 
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_scale_group.go
```diff
@@ -389,21 +389,20 @@ func resourceSoftLayerScaleGroupRead(d *schema.ResourceData, meta interface{}) e
...
 	}
-
 	virtualGuestTemplate := populateMemberTemplateResourceData(*slGroupObj.VirtualGuestMemberTemplate)
 	d.Set("virtual_guest_member_template", virtualGuestTemplate)
 
 	return nil
 }
 
-func populateMemberTemplateResourceData(template datatypes.Virtual_Guest) map[string]interface{} {
+func populateMemberTemplateResourceData(template datatypes.Virtual_Guest) []map[string]interface{} {
```
renier, line 405: why do you need to return an array of maps? didn't see where you could have more than one map below...

minsikl: `virtual_guest_member_template` field has `schema.TypeList` type.

PR #36, opened 2016-10-24
Feature/globalip

minsikl:
Add a global IP resource. README, Global IP resource, and test codes are added.

### DIFF
--- a/softlayer/resource_softlayer_global_ip.go
+++ b/softlayer/resource_softlayer_global_ip.go
@@ -114,13 +114,13 @@ func resourceSoftLayerGlobalIpUpdate(d *schema.ResourceData, meta interface{}) e
                 return fmt.Errorf("Not a valid global ip ID, must be an integer: %s", err)
         }
 
-        opts := datatypes.Network_Subnet_IpAddress_Global{}
+        //opts := datatypes.Network_Subnet_IpAddress_Global{}
 
         //if d.HasChange("routes_to") {
         //      opts.IpAddress.IpAddress = sl.String(d.Get("routes_to").(string))
         //}
 
-        _, err := service.Id(globalIpId).route(sl.String(d.Get("routes_to").(string)))
+        _, err = service.Id(globalIpId).Route(sl.String(d.Get("routes_to").(string)))
 
         if err != nil {
                 return fmt.Errorf("Error editing Global Ip: %s", err)
--- a/softlayer/resource_softlayer_global_ip.go
+++ b/softlayer/resource_softlayer_global_ip.go
@@ -105,6 +105,52 @@ func resourceSoftLayerGlobalIpRead(d *schema.ResourceData, meta interface{}) err
         return nil
 }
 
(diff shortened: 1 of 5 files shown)


softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,270 @@
...
+	log.Println("[INFO] Creating global ip")
+
+	receipt, err := services.GetProductOrderService(sess).
+		PlaceOrder(productOrderContainer, sl.Bool(false))
+	if err != nil {
+		return fmt.Errorf("Error during creation of global ip: %s", err)
+	}
+
+	globalIp, err := findGlobalIpByOrderId(sess, *receipt.OrderId)
+	d.SetId(fmt.Sprintf("%d", *globalIp.Id))
```
renier, line 78: Have to set "ip_address" too on Create, since Update won't set it later.

softlayer/resource_softlayer_global_ip.go
```diff
@@ -0,0 +1,270 @@
...
+	service := services.GetNetworkSubnetIpAddressGlobalService(sess)
+
+	globalIpId, err := strconv.Atoi(d.Id())
+	if err != nil {
+		return false, fmt.Errorf("Not a valid ID, must be an integer: %s", err)
+	}
+
+	result, err := service.Id(globalIpId).GetObject()
+	if err != nil {
+		return false, fmt.Errorf("Error retrieving global ip: %s", err)
```
renier, line 187: I think here we need to differentiate between an error with the api, and simply a 404, which would mean it does not exist (we would return `false, nil` in that case).

softlayer/resource_softlayer_global_ip_test.go
```diff
@@ -0,0 +1,155 @@
...
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: testAccCheckSoftLayerGlobalIpConfig_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckSoftLayerGlobalIpExists("softlayer_global_ip.test-global-ip"),
+					testAccCheckSoftLayerResources("softlayer_global_ip.test-global-ip", "routes_to",
+						"softlayer_virtual_guest.vm1", "ipv4_address"),
+				),
```
renier, line 29: I would add another check to ensure ip_address in the global ip resource is set using `resource.TestMatchResourceAttr()`

softlayer/resource_softlayer_global_ip.go
minsikl: Added

softlayer/resource_softlayer_global_ip.go
minsikl: Applied.

softlayer/resource_softlayer_global_ip_test.go
minsikl: Added the checking logic.

PR #42, opened 2016-10-26
Fix VPX REST status checking logic for VPX 10.5.

minsikl:
When a new netscaler VPX is created, it requires additional time for service initializing. The nitro API service status can be checked with NADCService.GetLoadBalancers() function as follows:

1) netscaler VPX 10.1
- When the nitro API service is not initialized : NADCService.GetLoadBalancers() returns an error :`{"error":"Could not connect to host","code":"HTTP"}` 
- When the nitro API service completes the initialization : NADCService.GetLoadBalancers() returns `nil`

2) netscaler VPX 10.5
- When the nitro API service is not initialized : NADCService.GetLoadBalancers() returns an error :`{"error":"Could not connect to host","code":"HTTP"}` 
- When the nitro API service completes the initialization : NADCService.GetLoadBalancers() returns `{"error":"There was a problem processing the reply from the application tier.  Please contact development.","code":"SoftLayer_Exception_Public"}`

### DIFF
--- a/softlayer/resource_softlayer_lb_vpx.go
+++ b/softlayer/resource_softlayer_lb_vpx.go
@@ -424,13 +424,15 @@ func resourceSoftLayerLbVpxCreate(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Failed to create VIPs for Netscaler VPX ID: %d", id)
 	}
 
-	// Wait VPX service initializing. GetLoadBalancers() internally calls REST API of VPX and returns an error
-	// if the REST API is not available.
+	// Wait VPX service initializing. GetLoadBalancers() internally calls REST API of VPX and returns
+	// an error "Could not connect to host" if the REST API is not available.
 	IsRESTReady := false
 
 	for restWaitCount := 0; restWaitCount < 60; restWaitCount++ {
 		_, err := NADCService.Id(id).GetLoadBalancers()
-		if err == nil {
+		// GetLoadBalancers returns an error "There was a problem processing the reply from the
+		// application tier.  Please contact development." if the VPX version is 10.5.
+		if err == nil || !strings.Contains(err.Error(), "Could not connect to host") {
 			IsRESTReady = true
 			break
 		}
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_lb_vpx.go
```diff
@@ -424,13 +424,15 @@ func resourceSoftLayerLbVpxCreate(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Failed to create VIPs for Netscaler VPX ID: %d", id)
 	}
 
-	// Wait VPX service initializing. GetLoadBalancers() internally calls REST API of VPX and returns an error
-	// if the REST API is not available.
+	// Wait VPX service initializing. GetLoadBalancers() internally calls REST API of VPX and returns
```
renier: should this say _Wait while VPX service is initializing_?

softlayer/resource_softlayer_lb_vpx.go
```diff
@@ -424,13 +424,15 @@ func resourceSoftLayerLbVpxCreate(d *schema.ResourceData, meta interface{}) erro
...
+	// an error "Could not connect to host" if the REST API is not available.
 	IsRESTReady := false
 
 	for restWaitCount := 0; restWaitCount < 60; restWaitCount++ {
 		_, err := NADCService.Id(id).GetLoadBalancers()
-		if err == nil {
+		// GetLoadBalancers returns an error "There was a problem processing the reply from the
+		// application tier.  Please contact development." if the VPX version is 10.5.
+		if err == nil || !strings.Contains(err.Error(), "Could not connect to host") {
```
renier, line 435: so if there is an error, we think REST is ready, if the error is not about not connecting to the host? Can there be other types of error that impair REST being ready?

minsikl: It looks like that SoftLayer API tries to connect to VPX with REST protocol, then checks the VPX version and returns `{"error":"There was a problem processing the reply from the application tier.  Please contact development.","code":"SoftLayer_Exception_Public"}` if the VPX version is not 10.1

renier: can we expand the condition to this? `err == nil || (vpxVersion != "10.1" && !strings.Contains(err.Error(), "Could not connect to host"))`

PR #52, opened 2016-11-16
Feature vpx ssl

minsikl:
Add SSL Offload function for netscaler VPX 10.5
softlayer_lb_vpx_vip supports "SSL" protocol and softlayer_security_certificate resource is used as a security certification.


renier:
@minsikl Does the vendor directory need to be updated with these changes?

minsikl:
@renier No, the vendor directory doesn't have to be updated.

PR #53, opened 2016-11-25
Use PlaceOrder instead of CreateObject for virtual guests with custom…

### LINKED ISSUE #50: [softlayer_virtual_guest] Cannot provision vm from custom image template with multiple disks
e.g. applying plan
```
resource "softlayer_virtual_guest" "vsi" {
    cores   = 1
    ram      = 1024
    disks    = [ 25, 10 ]
    name  = "test"
    domain = "domain.com"
    image_id = <custom image id>
    ssh_keys = [ "<some key>" ]
    datacenter = "osl01"
    local_disk = false
    hourly_billing = true
    private_network_only = true
}
```

Results in the following error:
```
Error applying plan:

1 error(s) occurred:

* softlayer_virtual_guest.vsi: Error creating virtual guest: SoftLayer_Exception_InvalidValue: Invalid value provided for 'blockDevices'. Block devices may not be provided when using an image template. (HTTP 500)
```

Further research shows that this is known issue:

- softlayer/softlayer-python/issues/693
- http://stackoverflow.com/questions/37128160/error-when-using-softlayer-ruby-api-to-specific-additional-disks-with-a-virtual

Any chance this will be supported through terraform using customOrder like described [here](http://sldn.softlayer.com/blog/bpotter/going-further-softlayer-api-python-client-part-3)?
Is there any other workaround that can be used in automated way (the only viable alternative right now is to manually order through Portal)?

matevzmihalic:
… image and multiple disks. Fixes #50

### DIFF
--- a/softlayer/resource_softlayer_virtual_guest.go
+++ b/softlayer/resource_softlayer_virtual_guest.go
@@ -470,12 +470,62 @@ func resourceSoftLayerVirtualGuestCreate(d *schema.ResourceData, meta interface{
 
 	log.Println("[INFO] Creating virtual machine")
 
-	guest, err := service.CreateObject(&opts)
-	if err != nil {
-		return fmt.Errorf("Error creating virtual guest: %s", err)
+	var id int
+
+	if opts.BlockDevices != nil && opts.BlockDeviceTemplateGroup != nil {
+		bd := *opts.BlockDeviceTemplateGroup
+		opts.BlockDeviceTemplateGroup = nil
+
+		// Find a temporary OS reference code that will be used to generate a template
+		tgService := services.GetVirtualGuestBlockDeviceTemplateGroupService(meta.(ProviderConfig).SoftLayerSession())
+		descriptions, err := tgService.GetVhdImportSoftwareDescriptions()
+		if err != nil {
+			return fmt.Errorf("Error getting vhdi descriptions: %s", err)
+		}
+		var referenceCode *string
+		for _, d := range descriptions {
+			if *d.Name == "Ubuntu" {
+				referenceCode = d.ReferenceCode
+				break
+			}
+		}
+
+		opts.OperatingSystemReferenceCode = referenceCode
+		template, err := service.GenerateOrderTemplate(&opts)
+		if err != nil {
+			return fmt.Errorf("Error generating order template: %s", err)
+		}
+
+		// Remove temporary OS from actual order
+		for i, p := range template.Prices {
+			if strings.Contains(*p.Item.Description, "Ubuntu") {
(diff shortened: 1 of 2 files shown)


renier:
@matevzmihalic Are you not able to generate the order template without an os reference code but with the image id instead (or with the image global identifier)?

matevzmihalic:
If I use image id and multiple disks to generate the order template I get the same error as before (`Invalid value provided for 'blockDevices'. Block devices may not be provided when using an image template`).

softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -470,12 +470,62 @@ func resourceSoftLayerVirtualGuestCreate(d *schema.ResourceData, meta interface{
...
-	guest, err := service.CreateObject(&opts)
-	if err != nil {
-		return fmt.Errorf("Error creating virtual guest: %s", err)
+	var id int
+
+	if opts.BlockDevices != nil && opts.BlockDeviceTemplateGroup != nil {
+		bd := *opts.BlockDeviceTemplateGroup
+		opts.BlockDeviceTemplateGroup = nil
+
+		// Find a temporary OS reference code that will be used to generate a template
```
renier: You can use something like "UBUNTU_LATEST" which is a code that won't be going away any time soon.

renier (changes requested):

softlayer/resource_softlayer_virtual_guest.go
matevzmihalic: Done.

softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -470,12 +470,47 @@ func resourceSoftLayerVirtualGuestCreate(d *schema.ResourceData, meta interface{
...
+		opts.OperatingSystemReferenceCode = sl.String("UBUNTU_LATEST")
+		template, err := service.GenerateOrderTemplate(&opts)
+		if err != nil {
+			return fmt.Errorf("Error generating order template: %s", err)
+		}
+
+		// Remove temporary OS from actual order
+		for i, p := range template.Prices {
+			if strings.Contains(*p.Item.Description, "Ubuntu") {
+				template.Prices = append(template.Prices[:i], template.Prices[i+1:]...)
```
renier: this continual stitching of the array may result in a high-ish number of memory allocations.

what if you create a separate price array that has all the prices whose item description do not contain the world ubuntu. then you set it `template.Prices = ubuntuLessPricess`

renier (changes requested):

softlayer/resource_softlayer_virtual_guest.go
matevzmihalic: Done.

Issue #58, opened 2016-12-06
network_speed value is not updated for virtual guest with private network only


minsikl:
If `softlayer_virtual_guest` resource definition has `private_network_only = true` but doesn't have `network_speed`, `terraform show` returns 0 for network_speed.

Example: test.tf
```
resource "softlayer_virtual_guest" "vmi" {
    cores   = 1
    memory      = 1024
    disks    = [ 25 ]
    hostname  = "test"
    domain = "example.com"
    os_reference_code = "DEBIAN_7_64"
    datacenter = "osl01"
    local_disk = false
    hourly_billing = true
    private_network_only = true
}
```

```
# terraform show
...
  ip_address_id_private = 66541905
  ipv4_address_private = 10.180.16.199
  local_disk = false
  memory = 1024
  network_speed = 0
...
```

PR #59, opened 2016-12-06
Change network speed reference for network_speed in virtual_guest

minsikl:
If a virtual guest only has private network interface, PrimaryNetworkComponent.MaxSpeed returns 0.
To support network_speed for the virtual guest with private network only, I changed network
 speed reference from PrimaryNetworkComponent.MaxSpeed to PrimaryBackendNetworkComponent.MaxSpeed.

This PR resolves the issue #58

### DIFF
--- a/softlayer/resource_softlayer_virtual_guest.go
+++ b/softlayer/resource_softlayer_virtual_guest.go
@@ -577,7 +577,7 @@ func resourceSoftLayerVirtualGuestRead(d *schema.ResourceData, meta interface{})
 		"network_speed",
 		sl.Grab(
 			result,
-			"PrimaryNetworkComponent.MaxSpeed",
+			"PrimaryBackendNetworkComponent.MaxSpeed",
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -577,7 +577,7 @@ func resourceSoftLayerVirtualGuestRead(d *schema.ResourceData, meta interface{})
 		"network_speed",
 		sl.Grab(
 			result,
-			"PrimaryNetworkComponent.MaxSpeed",
+			"PrimaryBackendNetworkComponent.MaxSpeed",
```
renier, line 580: Will this still report the correct network speed for public VMs?

minsikl: Yes, it reports the correct MaxSpeed for both public VMs and private VMs.

PR #60, opened 2016-12-07
Manually set UserData for virtual guests

matevzmihalic:
Because `GenerateOrderTemplate` doesn't return `UserData` we have to manually set it from `opts`.

I think this is a bug in SoftLayer API or `softlayer-go`, but until it's fixed we have to use this workaround.


softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -496,6 +496,9 @@ func resourceSoftLayerVirtualGuestCreate(d *schema.ResourceData, meta interface{
 		template.VirtualGuests[0].BlockDeviceTemplateGroup = &bd
 		template.VirtualGuests[0].OperatingSystemReferenceCode = nil
 
+		// GenerateOrderTemplate ommits UserData, so we set it manually
```
renier: typo in comment. should be 'omits'.

matevzmihalic: Fixed.

Issue #15, opened 2016-10-05
Support virtual guests with IPV6


fuzzyami:
We would like to spin up lots and lots of hosts with IPV6. The feature is woefully missing from the original terraform plugin for SL. It is supported in this version?

thanks!

renier:
@fuzzyami This is not available yet.

fuzzyami:
Thanks for the prompt reply. This seriously inhibits our ability to deploy hosts at large scale.

minsikl:
SoftLayer vm supports an IPv6 address for public network. The following arguments can be added to a schema of a virtual guest.
```
			"ipv6_address": {
				Type:     schema.TypeString,
				Computed: true,
			},

			"ip_address_id6": {
				Type:     schema.TypeInt,
				Computed: true,
			},

			"public_subnet6": {
				Type:     schema.TypeString,
				Optional: true,
				ForceNew: true,
				Computed: true,
			},
```

minsikl:
Public IPv6 address feature has been added to virtual guests and the issue can be closed.

PR #63, opened 2016-12-09
Support virtual guests with public IPv6 addresses

minsikl:
Add IPv6 features to a virtual guest resource and fix minor bugs as follows: 
- Add ipv6_enabled, ipv6_address, ip_address_id6, and public_subnet6 to a virtual_guest schema.
- Remove createObject call and use PlaceOrder for virtual_guest creation.
- Fix user_metadata setting bug.

This PR provides function for #15.

### DIFF
--- a/docs/resources/softlayer_virtual_guest.md
+++ b/docs/resources/softlayer_virtual_guest.md
@@ -139,7 +139,7 @@ The following attributes are exported:
 * `ipv4_address_private` - Private IPv4 address of the virtual guest.
 * `ip_address_id` - Unique ID for the public IPv4 address assigned to the virtual_guest.
 * `ipv6_address` - Public IPv6 address of the virtual guest. It is provided when `ipv6_enabled` is `true`.
-* `ip_address_id6` - Unique ID for the public IPv6 address assigned to the virtual_guest. It is provided when `ipv6_enabled` is `true`.
-* `public_subnet6` - Public IPv6 subnet. It is provided when `ipv6_enabled` is `true`.
+* `ipv6_address_id` - Unique ID for the public IPv6 address assigned to the virtual_guest. It is provided when `ipv6_enabled` is `true`.
+* `public_ipv6_subnet` - Public IPv6 subnet. It is provided when `ipv6_enabled` is `true`.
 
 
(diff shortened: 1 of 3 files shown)


docs/resources/softlayer_virtual_guest.md
```diff
@@ -124,13 +124,22 @@ The following arguments are supported:
...
 
 * `id` - id of the virtual guest.
 * `ipv4_address` - Public IPv4 address of the virtual guest.
-* `ip_address_id_private` - Unique ID for the private ID address assigned to the virtual_guest.
+* `ip_address_id_private` - Unique ID for the private IPv4 address assigned to the virtual_guest.
 * `ipv4_address_private` - Private IPv4 address of the virtual guest.
-* `ip_address_id` - Unique ID for the public ID address assigned to the virtual_guest.
+* `ip_address_id` - Unique ID for the public IPv4 address assigned to the virtual_guest.
+* `ipv6_address` - Public IPv6 address of the virtual guest. It is provided when `ipv6_enabled` is `true`.
+* `ip_address_id6` - Unique ID for the public IPv6 address assigned to the virtual_guest. It is provided when `ipv6_enabled` is `true`.
```
renier: I would call this `ipv6_address_id`

docs/resources/softlayer_virtual_guest.md
```diff
@@ -124,13 +124,22 @@ The following arguments are supported:
...
+* `public_subnet6` - Public IPv6 subnet. It is provided when `ipv6_enabled` is `true`.
```
renier: I would call this `public_ipv6_subnet`

renier (commented):
Should we also wait for the ipv6 address to be set when we wait for the VM to come up?

docs/resources/softlayer_virtual_guest.md
minsikl: done.

docs/resources/softlayer_virtual_guest.md
minsikl: done.

minsikl:
It looks like that we don't have to wait for the ipv6 address when we create a new VM.

Issue #65, opened 2016-12-12
Support multiple IPv4 addresses


andrew-morris:
I'm not sure if this is a planned feature or something that already exists but it would be great if there was an exposed terraform config to allow adding multiple NICs to a node.

minsikl:
SoftLayer vm provides one public network interface and one private network interface. If `private_network_only` is false in terraform configuration, the vm will have two network interfaces. Otherwise, the vm will have one private network interface.

renier:
@minsikl Can you order a VM with multiple IPs from the portal? or can that only be done later with IP blocks?

minsikl:
SoftLayer VM provides one private NIC and one public NIC. It is not possible to add additional virtual NICs to SoftLayer VMs. 

However, SoftLayer provides secondary IP addresses for VMs. They are public static IP addresses and can be ordered in VM creation menu. Once they are ordered together, a new VM will have secondary IP addresses as sub interfaces. Here is a result of secondary IP addresses:
```
eth0      Link encap:Ethernet  HWaddr 06:3e:59:97:1e:a4
          inet addr:10.86.107.133  Bcast:10.86.107.191  Mask:255.255.255.192
...

eth1      Link encap:Ethernet  HWaddr 06:70:fd:c5:a8:de
          inet addr:75.126.165.178  Bcast:75.126.165.191  Mask:255.255.255.240
...

eth1:0    Link encap:Ethernet  HWaddr 06:70:fd:c5:a8:de
          inet addr:75.126.14.4  Bcast:75.126.14.4  Mask:255.255.255.255
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1

eth1:1    Link encap:Ethernet  HWaddr 06:70:fd:c5:a8:de
          inet addr:75.126.14.5  Bcast:75.126.14.5  Mask:255.255.255.255
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1

eth1:2    Link encap:Ethernet  HWaddr 06:70:fd:c5:a8:de
          inet addr:75.126.14.6  Bcast:75.126.14.6  Mask:255.255.255.255
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1

eth1:3    Link encap:Ethernet  HWaddr 06:70:fd:c5:a8:de
          inet addr:75.126.14.7  Bcast:75.126.14.7  Mask:255.255.255.255
[...]

PR #67, opened 2016-12-15
Support secondary public IP addresses for virtual guest

minsikl:
- Added secondary_ip_count and secondary_ip_addresses arguments to virtual_guest.
- This PR will support functions for #65

### DIFF
--- a/docs/resources/softlayer_virtual_guest.md
+++ b/docs/resources/softlayer_virtual_guest.md
@@ -131,7 +131,6 @@ The following arguments are supported:
 *   `secondary_ip_count` | *int*
     * Provides secondary public IPv4 addresses. Acceptable values are 4 and 8. 
     * *Optional*
-    * *Default*: 0
 
 ## Attributes Reference
 
--- a/docs/resources/softlayer_virtual_guest.md
+++ b/docs/resources/softlayer_virtual_guest.md
@@ -129,7 +129,7 @@ The following arguments are supported:
     * *Optional*
     * *Default*: false
 *   `secondary_ip_count` | *int*
-    * Provides secondary public IPv4 addresses.
+    * Provides secondary public IPv4 addresses. Acceptable values are 4 and 8. 
     * *Optional*
     * *Default*: 0
 
--- a/docs/resources/softlayer_virtual_guest.md
+++ b/docs/resources/softlayer_virtual_guest.md
@@ -128,6 +128,10 @@ The following arguments are supported:
     * Provides a primary public IPv6 address.
     * *Optional*
     * *Default*: false
+*   `secondary_ip_count` | *int*
+    * Provides secondary public IPv4 addresses.
+    * *Optional*
+    * *Default*: 0
 
 ## Attributes Reference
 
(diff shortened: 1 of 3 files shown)


docs/resources/softlayer_virtual_guest.md
```diff
@@ -128,6 +128,10 @@ The following arguments are supported:
     * Provides a primary public IPv6 address.
     * *Optional*
     * *Default*: false
+*   `secondary_ip_count` | *int*
+    * Provides secondary public IPv4 addresses. Acceptable values are 4 and 8. 
+    * *Optional*
+    * *Default*: 0
```
renier: I think it is a little confusing saying that the default is 0 when acceptable values are 4 and 8. We should keep that to ourselves, and not say there is any default.

softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -270,6 +270,19 @@ func resourceSoftLayerVirtualGuest() *schema.Resource {
 				Computed: true,
 			},
 
+			"secondary_ip_count": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+				Default:  0,
```
renier: int type doesn't need to specify default of 0 I believe. That is the type's default already.

softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -718,6 +765,29 @@ func resourceSoftLayerVirtualGuestRead(d *schema.ResourceData, meta interface{})
...
+	if result.PrimaryIpAddress != nil {
+		secondarySubnetResult, err := services.GetAccountService(meta.(ProviderConfig).SoftLayerSession()).
+			Mask("ipAddresses[id,ipAddress]").
+			Filter(filter.Build(filter.Path("publicSubnets.endPointIpAddress.ipAddress").Eq(*result.PrimaryIpAddress))).
+			GetPublicSubnets()
+		if err != nil {
+			log.Printf("Error getting secondary Ip addresses: %s", err)
+		}
+
+		secondaryIps := make([]string, 0)
```
renier, line 778: if we know the number of secondary ips ahead of time, we can set the capacity in this make statement. `make([]string, 0, len(secondarySubnetResult))`. It is a small optimization.

minsikl: To get the number of secondary ips, I need to execute for loop and count it as follows:
```
		numberOfSecondaryIps := 0
		for _, subnet := range secondarySubnetResult {
			numberOfSecondaryIps += len(subnet)
		}
		secondaryIps := make([]string, 0, numberOfSecondaryIps)
```
It's difficult to expect optimization.

softlayer/resource_softlayer_virtual_guest.go
minsikl: done.

docs/resources/softlayer_virtual_guest.md
minsikl: done.

softlayer/resource_softlayer_virtual_guest.go
renier: ok, I see

PR #68, opened 2016-12-21
Making targetIP an optional parameter

ivan-n16:
- Initialized `ResourceRecords` with an empty struct when target is not given
- Updated readme

### DIFF
--- a/softlayer/resource_softlayer_dns_domain.go
+++ b/softlayer/resource_softlayer_dns_domain.go
@@ -44,7 +44,7 @@ func resourceSoftLayerDnsDomain() *schema.Resource {
 
 			"target": {
 				Type:     schema.TypeString,
-				Required: true,
+				Optional: true,
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


softlayer/resource_softlayer_dns_domain.go
```diff
@@ -59,14 +59,17 @@ func resourceSoftLayerDnsDomainCreate(d *schema.ResourceData, meta interface{})
...
+		opts.ResourceRecords = []datatypes.Dns_Domain_ResourceRecord{
+			{
+				Data: sl.String(d.Get("target").(string)),
+				Host: sl.String("@"),
+				Ttl:  sl.Int(86400),
+				Type: sl.String("a"),
+			},
+		}
+	} else {
+		opts.ResourceRecords = []datatypes.Dns_Domain_ResourceRecord{}
```
renier, line 72: Isn't this the default though for the Dns_Domain struct? an empty array.

ivan-n16: The default would be an empty array.
However, `json.Marshal` in `softlayer-go/session/rest.go` converts empty arrays to `null` 

Excerpt from https://golang.org/pkg/encoding/json/
```
Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value
```

I tested behavior with the following sample script,
```
package main

import (
        "fmt"
        "encoding/json"
)

type OuterStruct struct {
        Name string
        Addr []InnerStruct
}

type InnerStruct struct {
        Address string
}

func main() {
        a := OuterStruct{Name: "Ivan"}
        jsonVar1, _ := json.Marshal(a)
        fmt.Printf("%s\n", jsonVar1)

        a.Addr = []InnerStruct{}
        jsonVar2, _ := json.Marshal(a)
        fmt.Printf("%s\n", jsonVar2)

}
```

Output:
```
root@ubuntuEP01:~/terraformExamples/SlTerraExample# go run nestedStructTrial.go
{"Name":"Ivan","Addr":null}
{"Name":"Ivan","Addr":[]}
```

Issue #64, opened 2016-12-10
softlayer_virtual_guest.myapp-primary: unexpected EOF


Freyert:
### What was I doing?
Trying to provision some VMs.

### Versions
Terraform: v0.7.13
Go: go version go1.7.3 darwin/amd64

### Crash Log
```
Error applying plan:

3 error(s) occurred:

* softlayer_virtual_guest.myapp-primary: unexpected EOF
* softlayer_virtual_guest.myapp-secondary.1: unexpected EOF
* softlayer_virtual_guest.myapp-secondary.0: unexpected EOF

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
panic: runtime error: invalid memory address or nil pointer dereference
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc000c]
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64:
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64: goroutine 100 [running]:
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64: panic(0x73fd80, 0xc42000c0f0)
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64:     /usr/local/go/src/runtime/panic.go:500 +0x1a1
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64: github.com/softlayer/terraform-provider-softlayer/softlayer.resourceSoftLayerVirtualGuestRead(0xc4205c6060, 0x726d20, 0xc420343260, 0xc42057aa77, 0xc420637168)
2016/12/09 19:10:36 [DEBUG] plugin: terraform-provider-softlayer_darwin_amd64:     /Users/renier/Projects/go/src/github.com/softlayer/terraform-provider-softlayer/softlayer/resource_softlayer_virtual_guest.go:556 +0x6ac
[...]

PR #77, opened 2017-01-10
Fix virtual guest active transaction

minsikl:
Add a new function `WaitForVirtualGuestAvailable` to check the status of an active transaction, primary IP address, and secondary IP address at the same time. This fix will resolve the issue #64.

### DIFF
--- a/softlayer/resource_softlayer_virtual_guest.go
+++ b/softlayer/resource_softlayer_virtual_guest.go
@@ -919,78 +919,6 @@ func WaitForUpgradeTransactionsToAppear(d *schema.ResourceData, meta interface{}
 	return stateConf.WaitForState()
 }
 
-// WaitForIPAvailable Wait for the public ip to be available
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -977,61 +955,61 @@ func WaitForIPAvailable(d *schema.ResourceData, meta interface{}, private bool)
...
+			// Check active transactions.
+			log.Println("Checking active transactions.")
+			if result.ActiveTransaction != nil {
+				return result, "provisioning", nil
 			}
 
-			return transactions, "active", nil
+			// Check Primary IP address availability.
+			log.Println("Checking Primary IP address.")
+			if sl.Grab(result, ipAddress) == "" {
```
renier: sl.Get will perform better as it uses less reflection. Also, ipAddress does not happen to be a nested path.

minsikl:
applied.

PR #80, opened 2017-01-17
Fix virtual guest transactipon

minsikl:
`sl.Get(result, ipAddress)` does not return "" when the result doesn't have ipAddress.
Removed `sl.Get` from `WaitForVirtualGuestAvailable` and directly checked the value of `result. PrimaryBackendIpAddress` and `result.PrimaryIpAddress` This fix is related to the issue #64

### DIFF
--- a/softlayer/resource_softlayer_virtual_guest.go
+++ b/softlayer/resource_softlayer_virtual_guest.go
@@ -987,7 +987,6 @@ func WaitForVirtualGuestAvailable(d *schema.ResourceData, meta interface{}) (int
 
 			// Check Primary IP address availability.
 			log.Println("Checking primary backend IP address.")
-
 			if result.PrimaryBackendIpAddress == nil {
 				return result, "provisioning", nil
 			}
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -982,14 +979,20 @@ func WaitForVirtualGuestAvailable(d *schema.ResourceData, meta interface{}) (int
...
 			log.Println("Checking active transactions.")
 			if result.ActiveTransaction != nil {
 				return result, "provisioning", nil
 			}
 
 			// Check Primary IP address availability.
-			log.Println("Checking Primary IP address.")
-			if sl.Get(result, ipAddress) == "" {
+			log.Println("Checking primary backend IP address.")
+			if result.PrimaryBackendIpAddress == nil {
```
renier, line 990: `if sl.Get(result.PrimaryBackendIpAddress, "") == "" {`

I think that would have worked, and it also checks for a blank... although I don't know if a blank there is possible in the API response.

minsikl: result.PrimaryBackendIpAddress and result.PrimaryIpAddress have string pointer type and I don't have to use `sl.Get`.

renier: @minsikl So if string pointer is not nil the string will also not be blank?

minsikl: @renier Right, based on tests, the string pointer was nil or pointer of ipaddress string.

PR #101, opened 2017-02-21
Global ipv6 support

ivan-n16:
- Validate func checks if IPv6 is given in full form(to work around http://stackoverflow.com/questions/41250975/softlayer-api-global-ip-routing-not-working-with-ipv6)
- Price switching based on whether IPv4/IPv6 is ordered
- Removed comment mentioning support for only IPv4
- Doc update
- Check for validity of IPv4/IPv6
- Uncompress IPv6 if necessary

### DIFF
--- a/softlayer/resource_softlayer_global_ip.go
+++ b/softlayer/resource_softlayer_global_ip.go
@@ -127,13 +127,14 @@ func resourceSoftLayerGlobalIpUpdate(d *schema.ResourceData, meta interface{}) e
 
 	routes_to := d.Get("routes_to").(string)
 	if strings.Contains(routes_to, ":") && len(routes_to) != 39 {
-		if colonCount := strings.Count(routes_to, ":"); colonCount != 7 {
-			routes_to = strings.Replace(routes_to, "::", "::"+strings.Repeat(":", 7-colonCount), 1)
-		}
-
 		parts := strings.Split(routes_to, ":")
 		for x, s := range parts {
-			parts[x] = fmt.Sprintf("%04s", s)
+			if s == "" {
+				zeroes := 9 - len(parts)
+				parts[x] = strings.Repeat("0000:", zeroes)[:(zeroes*4)+(zeroes-1)]
+			} else {
+				parts[x] = fmt.Sprintf("%04s", s)
+			}
 		}
 
 		routes_to = strings.Join(parts, ":")
--- a/softlayer/resource_softlayer_global_ip.go
+++ b/softlayer/resource_softlayer_global_ip.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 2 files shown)


softlayer/resource_softlayer_global_ip.go
```diff
@@ -124,7 +125,22 @@ func resourceSoftLayerGlobalIpUpdate(d *schema.ResourceData, meta interface{}) e
...
+		parts := strings.Split(routes_to, ":")
+		for x, s := range parts {
+			parts[x] = fmt.Sprintf("%04s", s)
+		}
+
+		routes_to = strings.Join(parts, ":")
+		d.Set("routes_to", routes_to)
+	}
+
+	_, err = service.Id(globalIpId).Route(sl.String(routes_to))
```
renier, line 144: A couple of things:

1. We need to do this conversion for Create as well, no?
1. For Read, we should use net.ParseIP output and compare with net.ParseIP output of current value in `d`. Only if they are different, do we set `d` with new value (net.ParseIP output).

Good job in improving the expansion algorithm from the #golang channel. I would prefer though to do it this slightly faster way (replaces 3 calls to strings funcs with 1):

```go
        parts := strings.Split(routes_to, ":")
	for x, s := range parts {
		if s == "" {
                        zeroes := 9 - len(parts)
			parts[x] = strings.Repeat("0000:", zeroes)[:(zeroes*4)+(zeroes-1)]
		} else {
			parts[x] = fmt.Sprintf("%04s", s)
		}
	}
```

softlayer/resource_softlayer_global_ip.go
```diff
@@ -45,6 +47,16 @@ func resourceSoftLayerGlobalIp() *schema.Resource {
...
+					address := v.(string)
+					if net.ParseIP(address) == nil {
+						errors = append(errors, fmt.Errorf("Invalid IP"))
+					}
+					return
+				},
+				DiffSuppressFunc: func(k, o, n string, d *schema.ResourceData) bool {
+					return net.ParseIP(o).String() == net.ParseIP(n).String()
+				},
 			},
```
minsikl, line 60: In `DiffSuppressFunc`, the value `o` will have full ipv6 address format. However, `n` can have both full format or compressed format. So before compare `o` and `n`, it is necessary to uncompress `n`.

renier: `net.ParseIP()` will parse both compressed and uncompressed formats, and serialize (`.String()`) to the compressed format always.

https://play.golang.org/p/uASQqYtCpF

minsikl: Got it. Then the current code is acceptable.

softlayer/resource_softlayer_global_ip.go
ivan-n16: @renier 
1. Conversion will not be necessary in `resourceSoftLayerGlobalIpCreate` since `route_to` operation is not initiated there. `resourceSoftLayerGlobalIpCreate` calls `resourceSoftLayerGlobalIpUpdate` in the end wherein the `routes_to` is actually populated.
2. I did not understand. Are you suggesting update to `resourceSoftLayerGlobalIpRead` or to line 140 `d.Set("routes_to", routes_to)`?
3. Incorporated the modified code for decompressing IPv6 and tested it for creation/updation.

softlayer/resource_softlayer_global_ip.go
```diff
@@ -45,6 +47,16 @@ func resourceSoftLayerGlobalIp() *schema.Resource {
 			"routes_to": &schema.Schema{
 				Type:     schema.TypeString,
 				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					address := v.(string)
+					if net.ParseIP(address) == nil {
+						errors = append(errors, fmt.Errorf("Invalid IP"))
```
renier: Be more specific with the error `fmt.Errorf("Invalid IPv6 format: %s", address)`

Issue #113, opened 2017-02-27
Global IP acceptance test fails.


minsikl:
Global IP acceptance test fails as follows:
```
$ make testacc TEST=github.com/softlayer/terraform-provider-softlayer/softlayer TESTARGS="-run TestAccSoftLayerGlobalIp_Basic"
==> Checking that code complies with gofmt requirements...
go tool vet -all .
TF_ACC=1 go test github.com/softlayer/terraform-provider-softlayer/softlayer -v -run TestAccSoftLayerGlobalIp_Basic -timeout 120m
=== RUN   TestAccSoftLayerGlobalIp_Basic
[WARN] Test: Executing step 0
...
CREATE: softlayer_global_ip.test-global-ip
  ip_address: "" => "<computed>"
...
--- FAIL: TestAccSoftLayerGlobalIp_Basic (192.98s)
  testing.go:265: Step 0 error: Error applying: 1 error(s) occurred:

    * softlayer_global_ip.test-global-ip: diffs didn't match during apply. This is a bug with Terraform and should be reported as a GitHub Issue.

    Please include the following information in your report:

        Terraform Version: 0.8.5
        Resource ID: softlayer_global_ip.test-global-ip
        Mismatch reason: extra attributes: routes_to
        Diff One (usually from plan): *terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff{"ip_address":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}}, Destroy:false, DestroyDeposed:false, DestroyTainted:false}
        Diff Two (usually from apply): *terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff{"ip_address":*terraform.ResourceAttrDiff{Old:"", New:"", NewComputed:true, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}, "routes_to":*terraform.ResourceAttrDiff{Old:"", New:"192.155.227.105", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}}, Destroy:false, DestroyDeposed:false, DestroyTainted:false}

[...]

PR #114, opened 2017-02-27
Check if n has the appropriate IPv6 format in global IP.

minsikl:
When an interpolation syntax is used for _routes_to_, _net.ParseIP().String()_ function returns _nil_ and _DiffSuppressFunc()_ function returns _true_ in _resourceSoftLayerGlobalIpCreate()_ function.
In global IP acceptance test, _routes_to = "${softlayer_virtual_guest.vm2.ipv4_address}"_ and `terraform plan` ignores _routes_to_ as DiffSuppressFunc() returns `true`.
This fix checks if _routes_to_ value has the appropriate IPv6 format. 
It will resolve the issue #113

### DIFF
--- a/softlayer/resource_softlayer_global_ip.go
+++ b/softlayer/resource_softlayer_global_ip.go
@@ -55,7 +55,10 @@ func resourceSoftLayerGlobalIp() *schema.Resource {
 					return
 				},
 				DiffSuppressFunc: func(k, o, n string, d *schema.ResourceData) bool {
-					return net.ParseIP(o).String() == net.ParseIP(n).String()
+					newRoutesTo := net.ParseIP(n)
+					// Return true when n has the appropriate IPv6 format and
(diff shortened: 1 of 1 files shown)


renier:
@minsikl Are you saying that _DiffSuppressFunc_ gets the old/new values before interpolation has happened by Terraform?

minsikl:
@renier If interpolation is possible, _DiffSuppressFunc_ gets a new value after interpolation. However, if interpolation is not possible such as the following example, _DiffSuppressFun_ in `terraform plan` gets a new value before interpolation.  
```
resource "softlayer_virtual_guest" "vm2" {
    hostname = "vm2"
...
}

resource "softlayer_global_ip" "test-global-ip" {
    routes_to = "${softlayer_virtual_guest.vm2.ipv4_address}"
}
```

PR #117, opened 2017-03-01
IPv6 support for firewall rules

ivan-n16:
- Enables creation of IPv6 rules
- IPv6 expansion logic is not needed since API accepts compressed and uncompressed format


minsikl:
Looks good.

When a new firewall is created, it does not have any rule. However, once rules are added, at least one rule should remain. In delete function, default open rules for IPv4 are added. Can you add default open rules for IPv6?

PR #122, opened 2017-03-10
Block storage concurrency issue

ivan-n16:
- If `SoftLayer_Exception_Network_Storage_Group_MassAccessControlModification` is encountered, the code waits and retries.
- Block storage uses functions from file storage resource, hence change will only be required in `resource_softlayer_file_storage.go`.

### DIFF
--- a/softlayer/resource_softlayer_file_storage.go
+++ b/softlayer/resource_softlayer_file_storage.go
@@ -723,16 +723,24 @@ func updateAllowedIpAddresses(d *schema.ResourceData, sess *session.Session, sto
 			if len(ipObject) != 1 {
 				return fmt.Errorf("Number of IP address is %d", len(ipObject))
 			}
-			_, err = services.GetNetworkStorageService(sess).
-				Id(id).
-				AllowAccessFromHostList([]datatypes.Container_Network_Storage_Host{
-					{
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_file_storage.go
```diff
@@ -723,16 +723,24 @@ func updateAllowedIpAddresses(d *schema.ResourceData, sess *session.Session, sto
...
+					Id(id).
+					AllowAccessFromHostList([]datatypes.Container_Network_Storage_Host{
+						{
+							Id:         ipObject[0].Id,
+							ObjectType: sl.String("SoftLayer_Network_Subnet_IpAddress"),
+						},
+					})
+				if err != nil {
+					if strings.Contains(err.Error(), "SoftLayer_Exception_Network_Storage_Group_MassAccessControlModification") {
+						time.Sleep(5 * time.Second)
```
renier: This looks good. Let's make the time we wait between retries a constant `RetryTime` that can be used in all the relevant places.

ivan-n16: @renier Should `RetryTime` be made `retryTime`? So that it is not exported?

renier: yes, that's ok

PR #127, opened 2017-03-20
[Feature] Dns secondary resource

ivan-n16:
Facilitates creation, updation and deletion of a secondary DNS object


renier:
@ivan-n16 Hey, we are missing the documentation file under docs/resources

Issue #132, opened 2017-03-22
Error when using softlayer_file_storage resource


bonifaido:
terraform apply throws the following error when applying the following resource:
```
softlayer_file_storage.test: Creating...
  allowed_subnets.#:          "" => "1"
  allowed_subnets.xxxxxxxx: "" => "xxxxxxx/xx"
  capacity:                   "" => "20"
  datacenter:                 "" => "sjc01"
  hostname:                   "" => "<computed>"
  iops:                       "" => "4"
  snapshot_capacity:          "" => "20"
  type:                       "" => "Endurance"
  volumename:                 "" => "<computed>"
```

```
* softlayer_file_storage.test: Error while creating storage:No product items matching with keyName WRITEHEAVY_TIER and categoryCode storage_tier_level could be found
```

minsikl:
The following configuration works:
```
resource "softlayer_file_storage" "test" {
        type = "Endurance"
        datacenter = "sjc01"
        capacity = 20
        iops = 4
        snapshot_capacity = 20
}
```

Please post the configuration which generates the error.

bonifaido:
An example, that gives me a similar error message:
```
provider "softlayer" {}
resource "softlayer_file_storage" "storage-k8s-rabbitmq" {
  type       = "Performance"
  datacenter = "sjc01"
  capacity   = 20
  iops       = 300
}
```

```
softlayer_file_storage.storage-k8s-rabbitmq: Creating...
  capacity:   "" => "20"
  datacenter: "" => "sjc01"
  hostname:   "" => "<computed>"
  iops:       "" => "300"
  type:       "" => "Performance"
  volumename: "" => "<computed>"
softlayer_file_storage.storage-k8s-rabbitmq: Still creating... (10s elapsed)
Error applying plan:

1 error(s) occurred:

* softlayer_file_storage.storage-k8s-rabbitmq: 1 error(s) occurred:

* softlayer_file_storage.storage-k8s-rabbitmq: Error while creating storage:No product items matching with keyName 300_IOPS and categoryCode performance_storage_iops could be found
```

ivan-n16:
@minsikl Both `Endurance` and `Performance` configurations above are working for me. Perhaps this could be an issue of user not having adequate permissions?
As per [this](https://sldn.softlayer.com/article/permission-enforcement-softlayer-api):

[...]

PR #142, opened 2017-04-19
Fix storage resources with xml-rpc

minsikl:
- Applied latest xmlrpc and softlayer-go library which contain unmarshal error and timeout error fixes.
- Updated WaitForStorageAvailable to support xml-rpc.
- Check if price.LocationGroupId is 0 to support xml-rpc.
This fix resolves the issue #132

### DIFF
--- a/softlayer/resource_softlayer_file_storage.go
+++ b/softlayer/resource_softlayer_file_storage.go
@@ -594,17 +594,19 @@ func WaitForStorageAvailable(d *schema.ResourceData, meta interface{}) (interfac
 
 			// Check volume status.
 			log.Println("Checking volume status.")
-			resultStr := ""
+			var resultObj interface{}
 			err = sess.DoRequest(
 				"SoftLayer_Network_Storage",
 				"getObject",
 				nil,
 				&sl.Options{Id: &id, Mask: "volumeStatus"},
-				&resultStr,
+				&resultObj,
 			)
 			if err != nil {
 				return false, "retry", nil
 			}
+
+			resultStr := fmt.Sprintf("%v", resultObj)
 
 			if !strings.Contains(resultStr, "PROVISION_COMPLETED") &&
 				!strings.Contains(resultStr, "Volume Provisioning has completed") {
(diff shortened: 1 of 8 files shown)


softlayer/resource_softlayer_file_storage.go
```diff
@@ -594,18 +594,20 @@ func WaitForStorageAvailable(d *schema.ResourceData, meta interface{}) (interfac
...
-			resultStr := ""
+			var resultObj interface{}
 			err = sess.DoRequest(
 				"SoftLayer_Network_Storage",
 				"getObject",
 				nil,
 				&sl.Options{Id: &id, Mask: "volumeStatus"},
-				&resultStr,
+				&resultObj,
 			)
```
renier: Why not use `service.GetObject()` here? If there is an issue with this, we should fix down in softlayer-go.

minsikl: Please refer to the following link : http://stackoverflow.com/questions/41906025/softlayer-network-storage-volumestatus-is-not-a-string-type

renier: @minsikl This code works ok for me:
```go
package main

import (
	"fmt"
	"os"

	"github.com/softlayer/softlayer-go/services"
	"github.com/softlayer/softlayer-go/session"
)

func main() {
	sess := session.New()
	sess.Debug = true
	service := services.GetNetworkStorageService(sess)

	netStore, err := service.Id(22071867).Mask("volumeStatus").GetObject()
	if err != nil {
		fmt.Println(err)
		os.Exit(-1)
	}

	fmt.Println(*netStore.VolumeStatus)
}
// Prints "PROVISION_COMPLETED"
```

minsikl: @renier It mostly works. Hard to re-generate the problem. Some data center or some storage return different datatype than string type.

renier: @minsikl I saw that a problem existed when using XMLRPC, since the classic form object mask (used by default above) was not supported in our xmlrpc implementation and that is the only mask form that works in this case.

I have now fixed this in softlayer-go, so the code above will work whether using REST or XMLRPC. So we should simplify it in the provider now and update softlayer-go in the vendor subdirectory again.

minsikl: @renier It happens with both XML-RPC and REST API.

renier: @minsikl It will happen in any case then. However, using .GetObject as above results now in the same call you are doing using DoRequest, so it's not necessary to use DoRequest.

PR #144, opened 2017-04-25
Removing all tags from a Baremetal or Virtual Guest is done by using an empty string

jvrplmlmn:
This  PR should address https://github.com/softlayer/terraform-provider-softlayer/issues/139

Based on the documentation that can be found here: [https://github.com/softlayer/softlayer-go/blob/master/services/tag.go#L114](https://github.com/softlayer/softlayer-go/blob/58a39b003905ea63e4ad9bbfc48b3c9abfcea0bb/services/tag.go#L114)

Didn't had the chance to test it yet.


renier:
@jvrplmlmn Thanks.

@athak Could you try this out and tell us if it works for you?

Issue #43, opened 2016-10-28
Error message "Order is missing the following category: Disk Controller" when ordering bare-metal


giseongeom:
When tried to create bare-metal, terraform fail with the error 'Error ordering bare metal server: SoftLayer_Exception_Order_MissingCategory'
### example.tf

``` php
provider "softlayer" {
  username = "${var.sl_username}"
  api_key  = "${var.sl_api_key}"
}

# Create a new *Windows* bare metal
resource "softlayer_bare_metal" "sl_tf_sample2" {
  hostname             = "sl-tf-my2"
  domain               = "opsworld.io"
  os_reference_code    = "WIN_2012-STD-R2_64"
  datacenter           = "seo01"
  network_speed        = 1000
  hourly_billing       = true
  private_network_only = false
  fixed_config_preset = "D2620V4_64GB_2X1TB_SATA_RAID_1"
  tags = [
      "baremetal-guest"
    ]
}
```
### terraform plan and apply

```
$ terraform plan

+ softlayer_bare_metal.sl_tf_sample2
    datacenter:           "seo01"
    domain:               "opsworld.io"
    fixed_config_preset:  "D2620V4_64GB_2X1TB_SATA_RAID_1"
    hostname:             "sl-tf-my2"
    hourly_billing:       "true"
    network_speed:        "1000"
    os_reference_code:    "WIN_2012-STD-R2_64"
    private_ipv4_address: "<computed>"
    private_network_only: "false"
    private_subnet:       "<computed>"
    private_vlan_id:      "<computed>"
    public_ipv4_address:  "<computed>"
    public_subnet:        "<computed>"
    public_vlan_id:       "<computed>"
    tags.#:               "1"
    tags.3335598267:      "baremetal-guest"


Plan: 1 to add, 0 to change, 0 to destroy.

$ terraform apply

softlayer_bare_metal.sl_tf_sample2: Creating...


Error applying plan:

1 error(s) occurred:

* softlayer_bare_metal.sl_tf_sample2: Error ordering bare metal server: SoftLayer_Exception_Order_MissingCategory: Order is missing the following category: Disk Controller. (HTTP 500)

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```
### Environment Information

Operating System: msys on Windows 10 version 1607
Terraform version: 0.7.7 (Windows amd64)

[...]

PR #146, opened 2017-04-27
Update to pick up softlayer/softlayer-go#43

sykesm:
Pick up fix for JSON encoded strings.


renier:
This will happen once #142 is merged.

PR #151, opened 2017-05-26
Add resource to handle customized bare metal with quote

steven4ibm:
Actual bare metal resource handles only fixed configuration of bare metal.

This resource allows to manage customized bare metal where price and configuration are stored in IBM Bluemix Infrastructure (ex-SoftLayer) quote system.

### DIFF
--- a/softlayer/provider.go
+++ b/softlayer/provider.go
@@ -55,7 +55,6 @@ func Provider() terraform.ResourceProvider {
 		ResourcesMap: map[string]*schema.Resource{
 			"softlayer_virtual_guest":               resourceSoftLayerVirtualGuest(),
 			"softlayer_bare_metal":                  resourceSoftLayerBareMetal(),
-			"softlayer_bare_metal_quote":            resourceSoftLayerBareMetalQuote(),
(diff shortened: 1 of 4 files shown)


minsikl:
The plugin already has bare_metal resource. It is necessary to extend current bare_metal resource to support custom bare metal servers. 

Utilizing quote ID is a good idea and can simplify a resource definition and validation process. However, If the terraform resource definition does not contain specifications, it cannot be reused later. To fully support CRUD, it is necessary to have individual parameters in the terraform resource definition. One idea I'm thinking about is implementing a quote data source. Then, users can define custom bare metal server quotes on SL portal and import quote data sources. The quote data sources will retrieve bare metal server specifications, and users will define custom bare metal server with the information.

renier:
At a minimum, this would have to work as an additional parameter to the existing bare metal resource, and not as a separate resource.

I also like Minsik's idea of creating a data source for a bare metal quote. So in order for this to be merged, it would have to align with one of these two thoughts.

steven4ibm:
I have merge the bare_metal_quote resource with the bare_metal resource as requested. 

I don't have the Go skill neither the time to implement the second idea. But I am interested by seeing it implemented.

Issue #154, opened 2017-06-08
Provider Crashes when State file is out of sync


JimHart2:
The Softlayer provider errors with "Unexpected EOF" messages and then crashes when the state file thinks that a resource should exist that does not exist in the SoftLayer Environment.

minsikl:
Could you share sample tf and state files which crash the terraform?

JimHart2:
Attached files are the relevant parts of the config, the whole crash log, and the who state file.  This is a test environment. 

[crash.log.txt](https://github.com/softlayer/terraform-provider-softlayer/files/1062069/crash.log.txt)
[terraform.tfstate.txt](https://github.com/softlayer/terraform-provider-softlayer/files/1062070/terraform.tfstate.txt)
[users.tf.txt](https://github.com/softlayer/terraform-provider-softlayer/files/1062071/users.tf.txt)

renier:
@minsikl How are you reproducing this bug?

I tried `terraform refresh` and `terraform plan` using the provided files but it doesn't crash on the version of terraform indicated by the state file (i.e. 0.9.5). I also tried a cleaner state file with only the _btran_ user, but it didn't crash like that either.

minsikl:
@renier You can reproduce this bug with the following step:
1) Create a user using terraform
2) Delete a user in SL Portal
3) execute `terraform refresh`

When a user is deleted, it has remained with "CANCEL_PENDING" status in a specific period. Although a user ID still exists, the user information is already removed. When `terraform refresh` is executed, isExist returns `true` as the ID still exists, but Read function is crashed because some fields are already deleted.

PR #155, opened 2017-06-12
Checked a user with CANCEL_PENDING status in userExists function.

minsikl:
When a user is deleted, it has remained internally with "CANCEL_PENDING" status in a specific period. Although a user ID still exists, user information is already removed and cannot be recovered. The user with this status should be considered as a non-exist user. This fix will resolve the issue #154.

### DIFF
--- a/softlayer/resource_softlayer_user.go
+++ b/softlayer/resource_softlayer_user.go
@@ -472,6 +472,10 @@ func resourceSoftLayerUserExists(d *schema.ResourceData, meta interface{}) (bool
 
 	id, err := strconv.Atoi(d.Id())
 
+	if err != nil {
+		return false, fmt.Errorf("Not a valid ID, must be an integer: %s", err)
+	}
+
 	result, err := service.Id(id).GetObject()
 
 	// When a user is deleted, it has remained with "CANCEL_PENDING" status in specific time period.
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_user.go
```diff
@@ -474,6 +474,13 @@ func resourceSoftLayerUserExists(d *schema.ResourceData, meta interface{}) (bool
 
 	result, err := service.Id(id).GetObject()
 
+	// When a user is deleted, it has remained with "CANCEL_PENDING" status in specific time period.
+	// The user with this status should be considered as a non-exist user.
+	if result.UserStatus != nil && result.UserStatus.KeyName != nil &&
+		*result.UserStatus.KeyName == "CANCEL_PENDING" {
+		return false, nil
+	}
+
```
renier, line 487: Should we also check for a 404 error when using REST?

minsikl: Added codes to detect 404 error and other errors:
404 error -> return false, nil
Other errors(Such as authentication problem or network error) -> return false, Error

softlayer/resource_softlayer_user.go
```diff
@@ -481,7 +485,13 @@ func resourceSoftLayerUserExists(d *schema.ResourceData, meta interface{}) (bool
...
 	}
 
-	return result.Id != nil && *result.Id == id && err == nil, nil
+	if err != nil {
+		if apiErr, ok := err.(sl.Error); ok && apiErr.StatusCode == 404 {
+			return false, nil
+		}
+		return false, fmt.Errorf("Error retrieving user information: %s", err)
+	}
+	return true, nil
```
renier: as a precaution, you should still make sure you have a valid Id in the result and it matches the expected id.

`return result.Id != nil && *result.Id == id, nil`

Issue #147, opened 2017-05-02
Subnet Resource


tomwganem:
I have a use case where I need to order a portable subnet as part of my infrastructure, but there doesn't seem to be a resource for it. Is this something on the road map?

minsikl:
@tomwganem PR is opened for `softlayer_subnet` resource. It will support Portable and Static subnets. Please review the [specification](https://github.com/softlayer/terraform-provider-softlayer/blob/feature-subnet/docs/resources/softlayer_subnet.md).

PR #156, opened 2017-06-20
Feature subnet

minsikl:
`softlayer_subnet` resource supports portable subnet and static subnet management.
softlayer_subnet.md, resource_softlayer_subnet.go, and resource_softlayer_subnet_test.go files are added to implement `softlayer_subnet` resource. This PR will solve the issue #147

### DIFF
--- a/docs/resources/softlayer_subnet.md
+++ b/docs/resources/softlayer_subnet.md
@@ -26,7 +26,7 @@ The following example will create a private portable subnet which has one availa
 # Create a new portable subnet
 resource "softlayer_subnet" "portable_subnet" {
   type = "Portable"
-  network = "PRIVATE"
+  private = true
   ip_version = 4
   capacity = 4
   vlan_id = 1234567
@@ -41,7 +41,7 @@ The following example will create a public static subnet which has four availabl
 # Create a new static subnet
 resource "softlayer_subnet" "static_subnet" {
   type = "Static"
-  network = "PUBLIC"
+  private = false
   ip_version = 4
   capacity = 4
   endpoint_ip="151.1.1.1"
@@ -54,7 +54,7 @@ The following example returns first IP address in the subnet `test`:
 ```hcl
 resource "softlayer_subnet" "test" {
   type = "Static"
-  network = "PUBLIC"
+  private = false
   ip_version = 4
   capacity = 4
   endpoint_ip="159.8.181.82"
@@ -70,15 +70,16 @@ output "first_ip_address" {
 
 The following arguments are supported:
 
-* `network` | *string*
-    * Set the network property of the subnet if it is public or private. Accepted values are PRIVATE and PUBLIC.
+* `private` | *boolean*
+    * Set the network property of the subnet if it is public or private.
     * **Required**
 * `type` | *string*
     * Set the type of the subnet. Accepted values are Portable and Static.
     * **Required**
 * `ip_version` | *int*
     * Set the IP version of the subnet. Accepted values are 4 and 6.
-    * **Required**
+    * *Default*: true
+    * **Optional**
 * `capacity` | *int*
     * Set the size of the subnet.
     * Accepted values for a public portable IPv4 subnet are 4, 8, 16, and 32.
--- a/docs/resources/softlayer_subnet.md
+++ b/docs/resources/softlayer_subnet.md
@@ -107,4 +107,4 @@ The following arguments are supported:
 The following attributes are exported:
 
 * `id` - id of the subnet.
-* `subnet` - It rovides IP address/netmask format (ex. 10.10.10.10/28). It can be used to get an available IP address in `subnet`. 
+* `subnet_cidr` - It rovides IP address/cidr format (ex. 10.10.10.10/28). It can be used to get an available IP address in `subnet`. 
--- a/docs/resources/softlayer_subnet.md
+++ b/docs/resources/softlayer_subnet.md
@@ -84,8 +84,8 @@ The following arguments are supported:
     * Accepted values for a public portable IPv4 subnet are 4, 8, 16, and 32.
     * Accepted values for a private portable IPv4 subnet are 4, 8, 16, 32, and 64.
     * Accepted values for a public static IPv4 subnet are 1, 2, 4, 8, 16, and 32.
-    * Accepted value for a public portable IPv6 subnet is 64. 2^64 IP addresses will be provided.
-    * Accepted value for a public static IPv6 subnet is 64.  2^64 IP addresses will be provided.
+    * Accepted value for a public portable IPv6 subnet is 64. /64 block is created and 2^64 IP addresses are provided.
+    * Accepted value for a public static IPv6 subnet is 64.  /64 block is created and 2^64 IP addresses are provided.
     * **Required** 
 * `vlan_id` | *int*
     * VLAN id for portable subnet. It should be configured when the subnet is a portable subnet. Both public VLAN ID and private VLAN ID can 
--- a/docs/resources/softlayer_subnet.md
+++ b/docs/resources/softlayer_subnet.md
@@ -1,23 +1,25 @@
 #### `softlayer_subnet`
 
-Provides `Portable subnet` and `Static subnet` resource. Users are able to create `Public Portable Subnet`, `Private Portable Subnet`,
- and `Public Static Subnet` using `softlayer_subnet` resource.
+`softlayer_subnet` provides portable and static subnets that consist of either IPv4 and IPv6 addresses. Users are able to create 
+public portable subnets, private portable subnets, and public static subnets with an IPv4 option and public portable subnets and public static 
+subnets with an IPv6 option. 
  
-The `Portable subnet` is created as a seconday subnet in the VLAN and IP addresses in the subnet can be assigned as a secondary IP 
- address for resources in the VLAN. `Portable subnet` is an independant subnet and has a default gateway, network IP address, and broadcast 
- IP address. For example, if a subnet is `10.0.0.0/30`, `10.0.0.0` is a network IP address, `10.0.0.1` is a default gateway, and `10.0.0.3` 
- is a broadcast IP address. Therefore, only `10.0.0.2` can be assigned to resources as a secondary IP address. Number of usuable IP addresses is
- `capacity - 3`. If capacity is 4, number of usuable IP addresses is (4-3)=1. If capacity is 8, number of usuable IP addresses is (8-3)=5. 
- For additional details please refer to [Static and Portable IP blocks](https://knowledgelayer.softlayer.com/articles/static-and-portable-ip-blocks).
+The portable IPv4 subnet is created as a seconday subnet on a VLAN. IP addresses in the portable subnet can be assigned as secondary IP 
+ addresses for SoftLayer resources in the VLAN. Each portable subnet has a default gateway IP address, network IP address, and broadcast 
+ IP address. For example, if a portable subnet is `10.0.0.0/30`, `10.0.0.0` is a network IP address, `10.0.0.1` is a default gateway IP address, 
+ and `10.0.0.3` is a broadcast IP address. Therefore, only `10.0.0.2` can be assigned to SoftLayer resources as a secondary IP address. 
+ Number of usuable IP addresses is `capacity` - 3. If `capacity` is 4, the number of usuable IP addresses is 4 - 3 = 1. If `capacity` is 8, the 
+ number of usuable IP addresses is 8 - 3 = 5. For additional details, refer to [Static and Portable IP blocks](https://knowledgelayer.softlayer.com/articles/static-and-portable-ip-blocks).
 
-The `Static subnet` provides secondary IP addresses for a primary IP address. It works as a secondary IP address for a specific resource such as 
-virtual server, bare metal server, and netscaler VPX. Suppose that a virtual server requires additional secondary IP addresses. Then, users can create 
-`Static subnet` for the virtual server. Unlike `Portable subnet`, `capacity` is same with a number of usuable IP address. For example, if a subnet is 
-`10.0.0.0/30`, `10.0.0.0`~`10.0.0.3` are usuable IP addresses. 
- 
-For additional details please refer to [Subnet](https://knowledgelayer.softlayer.com/topic/subnets).
+The static IPv4 subnet provides secondary IP addresses for primary IP addresses. It provides secondary IP addresses for SoftLayer resources such as 
+virtual servers, bare metal servers, and netscaler VPXs. Suppose that a virtual server requires secondary IP addresses. Then, users can create 
+a static subnet on the public IP address of the virtual server. Unlike the portable subnet, `capacity` is same with a number of usuable IP address. 
+For example, if a static subnet is `10.0.0.0/30`, `capacity` is 4 and four IP addresses(10.0.0.0 ~ 10.0.0.3) can be used as secondary IP addresses. 
+For additional details, refer to [Subnet](https://knowledgelayer.softlayer.com/topic/subnets).
 
-The following example will create a private portable subnet which has one available portable IP address. 
+Both the public portable IPv6 subnet and the public static IP only accept `64` as a value of `capacity` attribute. They provide 2^64 IP addresses. For additional detail, refer to [IPv6 address](http://blog.softlayer.com/tag/ipv6)
+
+The following example will create a private portable subnet which has one available IPv4 address. 
 ##### Example Usage of portable subnet
 
 ```hcl
@@ -32,7 +34,7 @@ resource "softlayer_subnet" "portable_subnet" {
 }
 ```
 
-The following example will create a public static subnet which has four available portable IP address.
+The following example will create a public static subnet which has four available IPv4 address.
 ##### Example Usage of static subnet
 
 ```hcl
@@ -46,6 +48,24 @@ resource "softlayer_subnet" "static_subnet" {
   notes = "static_subnet_updated"
 }
 ```
+
+Sometimes, users need to get IP addresses on a subnet. Terraform built-in functions can be used to get IP addresses from `subnet`. 
+The following example returns first IP address in the subnet `test`:
+```hcl
+resource "softlayer_subnet" "test" {
+  type = "Static"
+  network = "PUBLIC"
+  ip_version = 4
+  capacity = 4
+  endpoint_ip="159.8.181.82"
+}
+
+# Use a built-in function cidrhost with index 0.
+output "first_ip_address" {
+  value = "${cidrhost(softlayer_subnet.test.subnet,0)}"
+}
+```
+
 ##### Argument Reference
 
 The following arguments are supported:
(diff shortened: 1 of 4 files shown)


docs/resources/softlayer_subnet.md
```diff
@@ -0,0 +1,109 @@
...
+
+Both the public portable IPv6 subnet and the public static IP only accept `64` as a value of `capacity` attribute. They provide 2^64 IP addresses. For additional detail, refer to [IPv6 address](http://blog.softlayer.com/tag/ipv6)
+
+The following example will create a private portable subnet which has one available IPv4 address. 
+##### Example Usage of portable subnet
+
+```hcl
+# Create a new portable subnet
+resource "softlayer_subnet" "portable_subnet" {
+  type = "Portable"
```
renier, line 28: if there can only be two types, portable or static, make it a boolean then `portable = true` or `static = true` to avoid possible spelling errors. If it is not portable, then it is static and vice-versa.

docs/resources/softlayer_subnet.md
```diff
@@ -0,0 +1,109 @@
...
+  network = "PRIVATE"
```
renier: same with the type of network. if there can only be two types, public and private, then make it a boolean `private = true`.

softlayer/resource_softlayer_subnet.go
```diff
@@ -0,0 +1,384 @@
...
+					if typeStr != "Portable" && typeStr != "Static" {
+						errs = append(errs, errors.New(
+							"type should be either Portable or Static."))
+					}
+					return
+				},
+			},
+
+			// IP version 4 or IP version 6
+			"ip_version": {
```
renier, line 71: could we have a default of `4` for the ip version?

softlayer/resource_softlayer_subnet.go
```diff
@@ -0,0 +1,384 @@
...
+			// endpoint_ip should be configured when type is "Static"
+			"endpoint_ip": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"vlan_id"},
+			},
+
+			// Provides IP address/netmask format (ex. 10.10.10.10/28)
+			"subnet": &schema.Schema{
```
renier: Do we have to validate the format here? Also, the resource is called subnet, so having a property also called the same can mislead the user. Can we calle it `cidr` or `cidr_mask`?

softlayer/resource_softlayer_subnet.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	d.Set("ip_version", *subnet.Version)
+	d.Set("capacity", *subnet.TotalIpAddresses)
+	if *subnet.Version == 6 {
+		d.Set("capacity", 64)
+	}
+	d.Set("subnet", *subnet.NetworkIdentifier+"/"+strconv.Itoa(*subnet.Cidr))
+	if subnet.Note != nil {
+		d.Set("notes", *subnet.Note)
+	}
```
renier: Not necessary, but these blocks could be refactored with `d.Set("notes", sl.Get(subnet.Note, nil))`

renier (changes requested):

docs/resources/softlayer_subnet.md
minsikl: SoftLayer internally uses some subnet types: `PRIMARY, ADDITIONAL_PRIMARY, SECONDARY, ROUTED_TO_VLAN, SECONDARY_ON_VLAN, STORAGE_NETWORK, and STATIC_IP_ROUTED`. From an end user perspective, the types can be simplified to: `Primary, Static, Portable, and Global`. We can ignore Global type because `softlayer_global_ip` is already provided. As of now, `softlayer_subnet` only supports `Static` and `Portable` because Primary subnet cannot be ordered manually. However, I think that Primary subnet ordering will be supported in the future. Then, we can extend this resource without schema update.

`softlayer_subnet` data source also can be added later. Unlike `softlayer_subnet` resource, the data source may supports `Primary, Portable, Static`. In this case, we can use the same attribute name `type`.

softlayer/resource_softlayer_subnet.go
minsikl: `subnet` is a computed attribute. So, validation is not necessary. In SoftLayer Portal, that column name is `subnet` and VLAN resource uses `subnet` as an attribute name of `IP address/cidr` format. I can change the attribute name to more user-friendly name. I think that `cidr` can be used as a name of this attribute. But, SoftLayer API uses this field name to represent `cidr block`(ex. /25). How about `ip_cidr` or `ip_address_cidr`?

renier: I see. If it is computed and SoftLayer portal calls it `subnet`, how about `subnet_cidr`?

docs/resources/softlayer_subnet.md
renier: @minsikl ok, in that case, why not use the same type strings as the SL API? What would `static` map to? `STATIC_IP_ROUTED`? What about `portable`?

The point is, if we can't make it fool-proof with booleans because there could be more than two types, then let's go with the raw API type names and then we can point there for the list of allowed types and spelling.

softlayer/resource_softlayer_subnet.go
minsikl: `virtual_guest`, `bare_metal`, `lb_vpx`, and `vlan` uses the name `subnet` or `public/private_sunbet` for this cidr notation. For consistency, I would like to use the name `subnet`. However `subnet_cidr` is also good.

docs/resources/softlayer_subnet.md
minsikl: @renier The initial code used `ROUTED_TO_VLAN` for portable subnets and `STATIC_IP_ROUTED` for static subnets. The subnet type can be retrieved using `SoftLayer_Network_Subnet ::subnetType`. However, I realized that SL API returns additional types such as `SECONDARY_ON_VLAN`, `SUBNET_ON_VLAN`, and `STATIC_IP_ROUTED`. After IPv6 feature is added, I got additional types such as `STATIC_IP6_ROUTED`. The raw subnet type is not an input parameter of a create function and I'm not able to guess which type will be returned from SL API. In end-user perspective, these raw types are meaningless and SL Portal uses terminology `Portable`, `Static`.

softlayer/resource_softlayer_subnet.go
renier: Yes, it should somehow say it is a cidr in the property. Explaining it as _"a subnet resource has a subnet property"_ is going to be a problem. For the other resources, that problem does not exist. Consistency with the other resources does not seem to be necessary for this.

docs/resources/softlayer_subnet.md
renier: Do we go with the portal or with the API? is the question.

If we go with the portal names, it would be good to at least document what that means in API types.

> The raw subnet type is not an input parameter of a create function and I'm not able to guess which type will be returned from SL API.

What kinds of subnets can you create then? and how do you specify the kind upon creation? (over the API)

minsikl: I want to go with API, but there are no proper subnet type codes in SL API. 
You can create only two types of subnet: Portable subnet and Static subnet using SL portal or SL API. Even if you create a portable/static subnet with same order template, the raw subnet type returned by SL API will be SECONDARY_ON_VLAN or SUBNET_ON_VLAN or ROUTED_TO_VLAN.

The PlaceOrder function requires package id, price item based on capacity, type, and IP version. If you want to create a portable IP, you should add EndPointVlanId parameter. For static IP, you should add EndPointIpAddressId parameter. For more detail, please refer to the function `buildSubnetProductOrderContainer`.

renier: Ok. If you provide vlan_id, it is portable, if you provide endpoint_ip, it is static. If that is how the API internally resolves the type of the subnet, is the type attribute still necessary then?

It seems like it would make sense to make it computed and just show the type that the platform decided on partially based on user input.

minsikl: @renier `softlayer_subnet` can be extended later to support different type of subnets such as `Primary`. If we don't have `type` attribute, we need to change the schema later to support different types. I don't think that `SoftLayer_Network_Subnet ::subnetType` have to be exposed to the end users. They are not used in a create function and don't contain any meaningful information. In the case of specification information such as load balancing method, heartbeat protocol, OS type, and Hardware type, SL API types provide clear key names and are used in resource creation process. However, the terminologies such as  'Portable', 'Private', and 'Primary' are used as offering names. SoftLayer portal, ticket, knowledge layer, and e-mail use them.

softlayer/resource_softlayer_subnet.go
minsikl: done

softlayer/resource_softlayer_subnet.go
minsikl: done

docs/resources/softlayer_subnet.md
minsikl: done

softlayer/resource_softlayer_subnet.go
minsikl: done

docs/resources/softlayer_subnet.md
renier: @minsikl Currently as I understand it, the `type` pseudo-attribute is not necessary to create a subnet to be portable or static. That distinction depends on whether other properties are set when you request a subnet creation. If in the future the SLAPI allows the creation of a primary network, we would have to wait to see how the user is expected to specify that. Will it be through an explicit property like a `type` or will it be by specifying some other kind of related property? We don't know and we should not guess at this point. However, I think we should represent the properties as-is in the resource and as they behave in the API and limit the number of pseudo-properties on the terraform side that we would have to keep in sync.

We could still create a pseudo-property `type` that is computed which will say "portable" or "static" based on the inferred type for the user's benefit. I just don't think it should be required.

minsikl: @renier `Potable` or `Static` are offering names. `Portable` subnet and `Static` subnet are different products and require different parameters. I used a term `softlayer_subnet` to generalize subnet products as much as possible. If you don't want to use this generalized name, you should create two different resources, `softlayer_portable_subnet`, `softlayer_static_subnet`.

minsikl: @renier SoftLayer already has APIs to manage Primary Subnets. SoftLayer just doesn't allow users to create new Primary Subnets manually at this time. Creating primary subnet requires same parameters with portable subnets.

softlayer/resource_softlayer_subnet.go
```diff
@@ -0,0 +1,386 @@
...
+	} else if *subnet.AddressSpace == "PUBLIC" {
+		d.Set("private", false)
+	}
+
+	d.Set("type", *subnet.SubnetType)
+	if strings.Contains(*subnet.SubnetType, "STATIC") {
+		d.Set("type", "Static")
+	} else if strings.Contains(*subnet.SubnetType, "VLAN") {
+		d.Set("type", "Portable")
+	}
```
renier: @minsikl There are `SubnetType`s that would not contain STATIC or VLAN. In that case "type" would be unset. Is that intentional?

docs/resources/softlayer_subnet.md
renier: @minsikl I'm good with the generalized subnet resource. What I'm saying now is that the `type` attribute should not be required. I think at the most, it should be computed. Let the users specify the required parameters for portable or static, depending on what they want, but they should not specify an additional type field that is not sent to the API upon create, which is just informational for the user.

minsikl: @renier Using the 'type' attribute has several advantages.

1. The subnet resource can support additional types without schema modification. Primary subnet attributes are the same as Portable subnet attributes. If the type attribute is a computed field, you will need to add the type attribute again to add Primary subnet later, which will create a situation where existing users will have to make configuration adjustments.

2. If the offering definition is separated by a parameter, you can create an unintended resource by user mistake. Users can make their intended subnet offerings without a mistake and provide more clarity.

minsikl: @renier It looks like that there is a misunderstanding in subnet ordering process. To get a price item for the ordering, you need to search KeyName such as `16_PORTABLE_PRIVATE_IP_ADDRESSES`. Package name also use `ADDITIONAL_SERVICES_STATIC_IP_ADDRESSES` and `ADDITIONAL_SERVICES_PORTABLE_IP_ADDRESSES`.  The terminology `Portable` and `Static` are internally used to distinguish subnet types.

softlayer/resource_softlayer_subnet.go
minsikl: @renier Portable subnets contain "VLAN" and Static subnets contain "STATIC". In the future, additional types can be added. For example, Primary subnets contain "PRIMARY".

renier: @minsikl Since there are some types that do not contain either word, technically, would it would be possible to get that from the API? If so, programmatically, it would be safer to add an else clause to handle that situation. For example, to throw an error in that case, since we don't support any other type of subnet.

docs/resources/softlayer_subnet.md
renier: @minsikl 

Maybe I'm not understanding something. 

>Using the 'type' attribute [...] The subnet resource can support additional types without schema modification.

How is this so?, since the data in the `type` attribute for the resource is not used other than to map it to the package type to get and to validate that the required resource attribute for that type was provided. However, the user could just provide the required resource attribute without specifying the type, and you could find the appropriate package type to search on based on this as well.

> If the offering definition is separated by a parameter, you can create an unintended resource by user mistake. Users can make their intended subnet offerings without a mistake and provide more clarity.

If a new subnet type is supported for creation that requires another attribute in the resource, we can add it. If it actually makes a type attribute necessary to differentiate we can add that too. If unset, for backwards compatibility, we can default based on the other required parameters provided (since we would already have that logic). But maybe another kind attribute would make more sense in that case (e.g. `primary = true/false`) rather than a type attribute. There is no way of knowing with certainty. Which is the folly of trying to code for an uncertain future. We should wait until it happens so that we can take a better look at how to handle that.

minsikl: > However, the user could just provide the required resource attribute without specifying the type,

I explained that primary subnet and portable subnet has the same attributes. How will you distinguish the primary subnet if you don't have type attribute?

> primary = true/false

`primary = true` is okay. But, what is `primary = false`? Is that mean that the subnet will be `global`, `secondary`, `third`, or `portable`? Will you add an attribute per types such as `static = true` and `secondary = true`? `primary` is one of subnet types. If you use `type` attribute, you don't have to add additional true/false attributes to support different types. Why do you use true/false for type names?

softlayer/resource_softlayer_subnet.go
minsikl: @renier Users may try to import different type of subnets. The current code will not generate an error. If you want to check the type strictly, you can throw an exception.

minsikl: Updated the code to generate an error when an invalid subnet type is detected.

docs/resources/softlayer_subnet.md
renier: >>However, the user could just provide the required resource attribute without specifying the type,

>I explained that primary subnet and portable subnet has the same attributes. How will you distinguish the primary subnet if you don't have type attribute?

Can a user order a primary subnet today? If not, then I stand by that we should not add a required pseudo-attribute based on something that has not happened in the API yet. We don't have enough information on how to handle until it actually does. Does this make sense?

Is it that we know that orderable primary subnets is in the works and we expect that to be available any moment now? If so, do we know exactly how the ordering of a primary subnet will look like?

minsikl: @renier `softlayer_vlan` already provides a function to create the first primary subnet. When you create a new VLAN using terraform, you should define the size of the first primary subnet. Then, `placeOrder` in `softlayer_vlan` resource sends priceId of VLAN and priceId of the primary subnet and creates a new VLAN with a primary subnet. So, primary subnets ordering requires same information with portable subnets ordering.

Subnet types are not pseudo-attributes. price item key names contain them, SL portal, automatic tickets, and knowledgelayer use them. Only SL API provides it's internal subnet types. 

Boolean attributes should be only used when SL API provides boolean properties or the attributes are optional values such as `advanced monitering` or `redundant power supply`. Or they will be pseudo-attributes, and it's hard to understand the meaning. 

I defined a general `softlayer_subnet` resource and it is a pseudo resource. That's why the `type` attribute is important in `softlayer_subnet` resource. `type` is added to define a real resource type of subnets. In the later, if a new attribute is added to describe a new subnet type such as `primary=true/false`, `primary` and `type` will provide duplicated information, and I don't think that it is a good design. Current `softlayer_subnet` only supports parts of SL subnet types. At least, we need to provide extensible schema if we want to use `softlayer_subnet`. 

If you don't want to use `type` attribute, let's just define separate resources `softlayer_portable_subnet` and `softlayer_static_subnet`. There is no advantage to using `softlayer_subnet` and if `softlayer_subnet` is provided without `type`, it will just make confusion.

renier: > softlayer_vlan already provides a function to create the first primary subnet. 

That's good to know and understand. I had not made the link between the vlan and the primary subnet. Why were you thinking that `primary` could be an additional type to `softlayer_subnet`, when it is already being provided by `softlayer_vlan`?

> Subnet types are not pseudo-attributes. price item key names contain them, SL portal, automatic tickets, and knowledgelayer use them. Only SL API provides it's internal subnet types.
>Boolean attributes should be only used when SL API provides boolean properties or the attributes are optional values such as advanced monitering or redundant power supply. Or they will be pseudo-attributes, and it's hard to understand the meaning.

In this case, the type is an attribute that is not synced with the API and is only managed locally. Thus, it is a pseudo-attribute of the resource by definition. Just as any boolean flags in other resources that are also only managed locally. The most common pseudo-attributes are booleans, because you usually don't need to make those kinds of attributes more complex and should not if possible.

> I defined a general softlayer_subnet resource and it is a pseudo resource.

Based on my definition above, a pseudo-resource would be something that is not sync-ed with the cloud and it would be managed completely locally. For example, the _random_ provider in terraform only has pseudo-resources.  It follows that `softlayer_subnet` is not a pseudo-resource.

> If you don't want to use type attribute, let's just define separate resources softlayer_portable_subnet and softlayer_static_subnet. There is no advantage to using softlayer_subnet and if softlayer_subnet is provided without type, it will just make confusion.

It doesn't have to be confusing if we document that portable subnets are created when you fill out the vlan_id, otherwise if you provide the endpoint_ip, it is static. Even less if you provide a documented computed attribute to the resource computing this fact for them as a word (portable or static).

Creating multiple resources, one per subnet type, is also an option, as long as 99% of the code is shared among them.

minsikl: >That's good to know and understand.

I gave you the primary subnet provisioning example as an answer to your question: `Can a user order a primary subnet today? If not, then I stand by that we should not add a required pseudo-attribute based on something that has not happened in the API yet. We don't have enough information on how to handle until it actually does.`. You can see that `primary subnet` requires same attributes with `portable subnet` when you create a new `primary subnet`.

> Why were you thinking that primary could be an additional type to softlayer_subnet, when it is already being provided by softlayer_vlan?

If you get a chance to deploy middle/large size systems or some cluster solutions on SoftLayer, you will find several `primary subnet` requirements which are not provided in SoftLayer. You can only create **the first primary subnet** as I explained above. You are not able to create 2nd/3rd primary subnets with a specific CIDR. Some cluster solutions require specific CIDR and some users want to reserve primary subnets in their VLAN for their firewall rules or applications.

> Based on my definition above, a pseudo-resource would be something that is not sync-ed with the cloud and it would be managed completely locally. For example, the random provider in terraform only has pseudo-resources. It follows that softlayer_subnet is not a pseudo-resource.

`primary subnet`, `portable subnet`, `static subnet`, and `global subnet` are combinations of interface and routing commands on physical routers and switches. `subnet` is a general network terminology which provides IP range and CIDR information. But, `primary subnet`, `portable subnet`, `static subnet`, and `global subnet` are not general terminologies and only used in SoftLayer. They are offering names and you can create and manage them on SoftLayer. `subnet` is used to refer to these soft layer offerings, not the actual resource name. 

> It doesn't have to be confusing if we document that portable subnets are created when you fill out the vlan_id, otherwise if you provide the endpoint_ip, it is static. Even less if you provide a documented computed attribute to the resource computing this fact for them as a word (portable or static).

Documents must, of course, be provided. Nevertheless, if there is a better way to define a resource, that method should be used.
Indirect resource definition methods also affect `tf` file readability. In a `tf` file where dozens of subnets are defined, people have to extract the subnet type by identifying the manual, deducing the subnet type, or creating a separate script.

> Creating multiple resources, one per subnet type, is also an option, as long as 99% of the code is shared among them.

I'll create create two different subnets `softlayer_subnet_portable` and `softlayer_subnet_static`.

PR #160, opened 2017-06-22
SoftLayer Virtual Guest: Include the postInstallScriptURI when importing the resource

jvrplmlmn:
Importing virtual guest resources to the Terraform state, adds them without the `post_install_script_uri` attribute. If the resource had that attribute defined, it will force a new resource creation.

I haven't worked with the Terraform code base that much, so I don't know how to verify the "import resource" scenario.

### DIFF
--- a/softlayer/resource_softlayer_virtual_guest.go
+++ b/softlayer/resource_softlayer_virtual_guest.go
@@ -670,7 +670,8 @@ func resourceSoftLayerVirtualGuestRead(d *schema.ResourceData, meta interface{})
 			"primaryVersion6IpAddressRecord[subnet,guestNetworkComponentBinding[ipAddressId]]," +
 			"primaryIpAddressRecord[subnet,guestNetworkComponentBinding[ipAddressId]]]," +
 			"primaryBackendNetworkComponent[networkVlan[id]," +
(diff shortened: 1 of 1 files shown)


softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -719,6 +720,7 @@ func resourceSoftLayerVirtualGuestRead(d *schema.ResourceData, meta interface{})
 	d.Set("private_network_only", *result.PrivateNetworkOnlyFlag)
 	d.Set("hourly_billing", *result.HourlyBillingFlag)
 	d.Set("local_disk", *result.LocalDiskFlag)
+	d.Set("post_install_script_uri", *result.PostInstallScriptUri)
```
minsikl, line 723: When resourceSoftLayerVirtualGuestRead function is invoked, `result.PostInstallScriptUri` returns `nil` and the terraform generates panic status. Please check if SL API returns `PostInstallScriptUri` properly. And post_install_script_url is an optional attribute. It is necessary to check if `result.PostInstallScriptUri` is `nil` or not in advance.

minsikl: SL API does not support PostInstallScriptUri. See the link : https://stackoverflow.com/questions/44728911/how-to-get-postinstallscripturi-information-of-softlayer-vitual-guest

PR #161, opened 2017-07-14
Feature hourly and quote based bare metal server

minsikl:
I cloned the PR #151 and extended the bare metal resource.

The following functions are added. 
Added hourly bare metal server provisioning functions.
* CPU, Memory, Disk and RAID configuration functions

Added quote based bare metal server provisioning functions.

Added quote_bare_metal data source.
* Data source which retrieves hourly bare metal server order information.


softlayer/resource_softlayer_bare_metal.go
```diff
@@ -76,100 +144,164 @@ func resourceSoftLayerBareMetal() *schema.Resource {
...
 
-			"user_metadata": {
-				Type:     schema.TypeString,
+			// Quote based provisioning only
+			"quote_id": {
+				Type:             schema.TypeInt,
+				Optional:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: applyOnce,
+			},
```
renier, line 263: I thought we were going to rely on information from the quote data source to fill in properties in the compute resource. Can you explain why we need the quote_id here?

minsikl: You can create a bare metal server with the quote_id attribute.

renier: Yes, so they can't use the data source for this? If not, what is the value-add of the quote data source?

minsikl: Quote data source provides required attributes for bare metal provisioning. You can refer to the quote data source to define a bare metal server resource. Or you can use interpolations to define a bare metal server resource. For example, 
```hcl
package_key_name = "${data.softlayer_quote_bare_metal.bm1.package_key_name}"
...
```

If you want to define a bare metal server resource with a quote data, you can write a definition as follows:
```hcl
data softlayer_quote_bare_metal quote_bm1{
  name = "quote1"
}

resource softlayer_bare_metal bm1 {
...
quote_id = "${data.softlayer_quote_bare_metal.quote_bm1.id}"
...
}
```

renier: I don't see why we would need both a quote id and a quote data source. Let's talk about that.

minsikl: Per your request, the quote_bare_metal data source was deleted.

PR #170, opened 2017-08-24
Add dedicated host option to virtual guest

matevzmihalic:
Added `dedicated_host_name` and `dedicated_host_id` options to virtual guest.

### DIFF
--- a/softlayer/resource_softlayer_virtual_guest.go
+++ b/softlayer/resource_softlayer_virtual_guest.go
@@ -145,9 +145,32 @@ func resourceSoftLayerVirtualGuest() *schema.Resource {
 			},
 
 			"dedicated_acct_host_only": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				ForceNew: true,
+				Type:          schema.TypeBool,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"dedicated_host_name", "dedicated_host_id"},
+			},
+
+			"dedicated_host_name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"dedicated_acct_host_only", "dedicated_host_id"},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					_, ok := d.GetOk("dedicated_host_id")
+					return new == "" && ok
+				},
+			},
+
+			"dedicated_host_id": {
+				Type:          schema.TypeInt,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"dedicated_acct_host_only", "dedicated_host_name"},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					_, ok := d.GetOk("dedicated_host_name")
+					return new == "0" && ok
+				},
 			},
 
 			"public_vlan_id": {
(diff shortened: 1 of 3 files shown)


softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -145,9 +145,32 @@ func resourceSoftLayerVirtualGuest() *schema.Resource {
...
+
+			"dedicated_host_name": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ConflictsWith: []string{"dedicated_acct_host_only", "dedicated_host_id"},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					_, ok := d.GetOk("dedicated_host_id")
+					return new == "" && ok
+				},
```
renier, line 162: Is this func necessary even with the conflictsWith?

softlayer/resource_softlayer_virtual_guest.go
```diff
@@ -417,6 +440,30 @@ func getVirtualGuestTemplateFromResourceData(d *schema.ResourceData, meta interf
...
+
+		if err != nil {
+			return opts, fmt.Errorf("Error looking up dedicated host '%s': %s", hostName, err)
+		} else if len(hosts) == 0 {
+			return opts, fmt.Errorf("Error looking up dedicated host '%s'", hostName)
+		}
+
+		opts.DedicatedHost = &hosts[0]
+	}
+
```
renier: given the restrictions in the schema, shouldn't these 3 if clauses be chained in else-if?

softlayer/resource_softlayer_virtual_guest.go
matevzmihalic: Yes it is needed. Because when you create VM terraforms saves `dedicated_host_name` and `dedicated_host_id` to `tfstate`. Then if you apply it again it gets empty string for `dedicated_host_name` from `*.tf` file and actual name from `tfstate`.
Without this function terraform recreates VM every time.

softlayer/resource_softlayer_virtual_guest.go
matevzmihalic: Because only one of those 3 options can be set each time we don't actually need chained if-else, but I can do it like that if you think it will make the code more readable.

matevzmihalic:
One thing I'm not sure about is if `dedicated_acct_host_only` and `dedicated_host_*` should be in conflict. Softlayer API allows you to to create VM with dedicated host id and `dedicated_acct_host_only` set to true or false although I think that it's logical to have only one of those set.

athak:
Could #172 be related to this? Is this a relatively recent API change?

softlayer/resource_softlayer_virtual_guest.go
renier: Yes, I think it would make it more explicit as to the mutual exclusivity we are trying to enforce for those options.

matevzmihalic: Fixed.

renier:
@matevzmihalic To make this complete, the only thing missing is to update the docs in `docs/resources/softlayer_virtual_guest.md`. Thanks.

matevzmihalic:
I added the docs.

c4po:
@renier @matevzmihalic 
This may not relate to this ticket. But I have trouble to provision vs on dedicate host. How can you get the dedicate_host_id from softlayer consul or slcli? I tried both dedicate_host_name and _id, terrform just cannot find it.

matevzmihalic:
@c4po last I checked (two months ago), the only way to create or get info about dedicated host was through web interface: [docs](https://console.bluemix.net/docs/vsi/vsi_provision_dedicated.html)

c4po:
@matevzmihalic how do you get the dedicate_host_id from softlayer web interface? the slcli tool doesn't provide the dedicate host list.

PR #174, opened 2017-08-26
Support configuring usip setting for vpx service

tomwganem:
Fix for #169


tomwganem:
Had to drop support for 10.1 because we aren't using the Nitro API.

This is now ready to be reviewed.

Issue #177, opened 2017-09-26
Build #489 (Add dedicated host option to virtual guest (#170)) failed


adriananeci:
It seems that the Build #489 failed most probably due to the old version of go.
I've also tried to build the project with make bin using Terraform v0.10.6 and go version go1.9 but still encountered some errors:
```
Errors:

  * provider.softlayer: Internal validation of the provider failed! This is always a bug
with the provider itself, and not a user issue. Please report
this bug:

19 error(s) occurred:

* resource softlayer_security_certificate: id is a reserved field name for a resource
* resource softlayer_scale_group: id is a reserved field name for a resource
* resource softlayer_basic_monitor: id is a reserved field name for a resource
* resource softlayer_block_storage: id is a reserved field name for a resource
* resource softlayer_dns_domain: id is a reserved field name for a resource
* resource softlayer_scale_policy: id is a reserved field name for a resource
* resource softlayer_lb_vpx_ha: id is a reserved field name for a resource
* resource softlayer_dns_secondary: id is a reserved field name for a resource
* resource softlayer_bare_metal: id is a reserved field name for a resource
* resource softlayer_ssh_key: id is a reserved field name for a resource
* resource softlayer_lb_vpx_service: id is a reserved field name for a resource
* resource softlayer_file_storage: id is a reserved field name for a resource
* resource softlayer_dns_domain_record: id is a reserved field name for a resource
* resource softlayer_user: id is a reserved field name for a resource
* resource softlayer_objectstorage_account: id is a reserved field name for a resource
* resource softlayer_virtual_guest: id is a reserved field name for a resource
* resource softlayer_provisioning_hook: id is a reserved field name for a resource
* resource softlayer_vlan: id is a reserved field name for a resource
* resource softlayer_global_ip: id is a reserved field name for a resource
```
Could be the terraform version which is causing these errors?

c4po:
Yes, terraform make the "id" a reserver filed name 
https://github.com/hashicorp/terraform/commit/67ceb1ab07acd55ebc0d7b6efb39019e73fd1888

PR #179, opened 2017-10-27
#177 remove id from resource schema

c4po:


renier:
A couple of issues with this:

1. We don't know which users are relying on this old behavior, so changing this may break them.
2. If we still decide to make this change to keep current with the latest terraform core, we should rename the id to something else (e.g. `domainId` for the domain resource) and update the documentation so as not to lose that information that might be needed by users (and would allow them to migrate to this new version more easily).

Thoughts?

c4po:
without make any change, this plugin won't work with new version of terraform. which force softlayer user to stay with old version of terraform. and we cannot get acc test pass.
once this changes merged in, we can bump up the release number so old version terraform user still can find the right version work for them,

I don't fell terraform user have access to the resource Id directly. If terraform can manage it though the plugin interface, why should we maintain another copy?

renier:
@c4po Yes, something should be done, however the SoftLayer Ids remain a problem. It is something that can certainly be accessed today. Seems to me we would just rename the Ids and update the docs.

renier:
#177

c4po:
Hi @renier , can you merge this PR and make a new release?

PR #180, opened 2017-10-29
replace id to softlater_id

c4po:
#177 
@renier

### DIFF
--- a/softlayer/data_source_softlayer_dns_domain.go
+++ b/softlayer/data_source_softlayer_dns_domain.go
@@ -13,7 +13,7 @@ func dataSourceSoftLayerDnsDomain() *schema.Resource {
 		Read: dataSourceSoftLayerDnsDomainRead,
 
 		Schema: map[string]*schema.Schema{
-			"id": &schema.Schema{
+			"softlayer_id": &schema.Schema{
 				Description: "A domain record's internal identifier",
 				Type:        schema.TypeInt,
 				Computed:    true,
--- a/softlayer/resource_softlayer_block_storage.go
+++ b/softlayer/resource_softlayer_block_storage.go
@@ -24,7 +24,7 @@ func resourceSoftLayerBlockStorage() *schema.Resource {
 		Importer: &schema.ResourceImporter{},
 
 		Schema: map[string]*schema.Schema{
-			"id": {
+			"softlayer_id": {
 				Type:     schema.TypeInt,
 				Computed: true,
 			},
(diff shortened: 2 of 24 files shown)


softlayer/resource_softlayer_block_storage.go
```diff
@@ -101,7 +101,7 @@ func resourceSoftLayerBlockStorage() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
-						"hostIQN": {
```
renier, line 104: was this change needed as well?

c4po: yes, this is required. There is a check in terraform that make sure all parameter name is lower case.

renier: Ok.

softlayer/data_source_softlayer_dns_domain.go
```diff
@@ -13,7 +13,7 @@ func dataSourceSoftLayerDnsDomain() *schema.Resource {
 		Read: dataSourceSoftLayerDnsDomainRead,
 
 		Schema: map[string]*schema.Schema{
-			"id": &schema.Schema{
```
renier, line 16: A bit confused about this still. Do we know if Terraform simply _owns_ the resource id, generating its own internal one, or is the intent for providers to never have to publish an id as part of its resource schema, but the expectation is still that the provider decides what the id should be? Maybe a question for the terraform mailing list?

If Terraform still expects the provider to decide what the resource id will be, but just does not want the id as part of the schema, then I understand now and we just need to remove the ids from the schema.

c4po: my understanding is Terraform will manage the resource id. That's why I created the first PR just remove them from the plugin.

renier: @c4po Sorry for my misunderstanding of this. I [asked on the Terraform mailing](https://groups.google.com/forum/#!topic/terraform-tool/8RSiwR86z60) list to make sure.

Terraform core actually does not manage the resource id. It is simply an implicit property. The provider will still assign the id to the resource. So the only change required is what you had done initially, which was to remove the id from the schema.

c4po: ok, do you want me close the PR and work on the previous one?

renier: Yes, what you had initially was enough (sorry :-/)

c4po: @renier is it ok to merge this and have a release

